PR #1, opened 2019-11-08
feat: add gh and npm tokens to travis.yml

abutterworth:


.travis.yml
```diff
@@ -1,12 +1,14 @@
 language: node_js
 node_js: 12
-before_install:
-- npm install -g npm@latest
-install:
-- npm install
```
davidjoy: We don't need to npm install?

PR #3, opened 2019-11-12
Adding config and pubSub interfaces

davidjoy:
Updating logging interface to conform with the others.

Also removed old logging methods that we don’t want anymore.


src/config/configure.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+
+let service = null;
+
+export default function configure(ConfigService, services) {
+  service = new ConfigService(services);
+  PropTypes.checkPropTypes({
```
abutterworth: Is this working?

src/logging/NewRelicLoggingService.js
```diff
@@ -28,98 +29,22 @@ class NewRelicLoggingService {
     }
   }
 
-  static logError(error, customAttributes) {
+  logError(error, customAttributes = {}) {
+    const errorCustomAttributes =
+      error.customAttributes !== undefined ? error.customAttributes : {};
```
abutterworth: Would `const errorCustomAttributes = error.customAttributes || {}` be safer?

src/config/configure.js
davidjoy: Wow.  Uh.  Sure do.  I failed to finish that code. ;)  Good catch.

src/logging/configure.js
```diff
@@ -0,0 +1,31 @@
+import PropTypes from 'prop-types';
+
+let service = null;
+
+export function configure(LoggingService, services) {
+  service = new LoggingService(services);
+
+  PropTypes.checkPropTypes({
```
abutterworth: Same thing here: is this a correct usage of checkPropTypes?

src/config/configure.js
abutterworth: Oh interesting you're using this to validate the implementation. Neat.

src/pubSub/PubSubJsService.js
```diff
@@ -0,0 +1,12 @@
+
+import PubSub from 'pubsub-js';
+
+export default class PubSubJsService {
```
abutterworth: What's the need for this?

PR #4, opened 2019-11-13
refactor: initial stab at a new initialize function.

davidjoy:
Now has explicit steps for configuring each service and then accepting (potentially) asynchronous handlers after configuration.  Only auth actually has a default handler.


src/base/handlers/auth.js
```diff
@@ -0,0 +1,23 @@
+import { getAuthenticatedApiClient, getAuthenticatedUser, ensureAuthenticatedUser } from '../../auth';
```
abutterworth: `getAuthenticatedApiClient` is now `getAuthenticatedHttpClient`

src/base/handlers/auth.js
```diff
@@ -0,0 +1,23 @@
...
+
+  if (requireAuthenticatedUser) {
+    authenticatedUser = await ensureAuthenticatedUser();
+  } else {
+    authenticatedUser = await getAuthenticatedUser();
+  }
+
+  if (authenticatedUser !== null && hydrateAuthenticatedUser) {
+    const { username } = authenticatedUser;
+    getAuthenticatedApiClient()
```
abutterworth: Is it worth a comment here that this does not use await in order to allow the rest of the app bootstrapping to proceed? (I assume this is the reason)

src/base/initialize.js
```diff
@@ -0,0 +1,89 @@
...
+  authService = null,
+  i18nService = null,
+  requireAuthenticatedUser = false,
+  hydrateAuthenticatedUser = false,
+  messages,
+  handlers = {},
+} = {}) {
+  try {
+    configurePubSub(pubSubService);
+    await override(() => {}, handlers.pubSub);
```
abutterworth: This override function is a little strange to me at first glance. What dropping it and using defaults instead – something like:
```
const handlers = Object.create(defaultHandlers, _handlers);

// ...
configureConfig(configService, { pubSubService: getPubSubService() });
handlers.config();
publish(APP_CONFIG_INITIALIZED);
```

src/base/handlers/auth.js
davidjoy: Good call.

src/base/initialize.js
davidjoy: Cool - I like that.

PR #5, opened 2019-11-13
feat: update analytics to new api usage

abutterworth:
exports `configure`, `SegmentAnalyticsService`, and all the existing analytics functions. Remove's an `initialize` function which is now executed internally upon calling `configure(impl, options)`.

New Usage:
```javascript 
configure(SegmentAnalyticsService, {
    loggingService,
    httpClient,
    analyticsApiBaseUrl,
    trackingLogApiBaseUrl: `${analyticsApiBaseUrl}/event`,
    apiKey: configuration.SEGMENT_KEY,
});
```


src/analytics/README.rst
```diff
@@ -15,19 +15,20 @@ To install frontend-analytics into your project::
...
     // Sample object containing app configuration
     import { configuration } from '../config/environment';
 
-    initializeSegment(configuration.SEGMENT_KEY);
-    configureAnalytics({
+    configureAnalytics(SegmentAnalyticsService, {
       loggingService: LoggingService,
-      authApiClient: apiClient,
+      httpClient: apiClient,
       analyticsApiBaseUrl: configuration.LMS_BASE_URL,
```
davidjoy: These config-based things can be snagged from the brand new ConfigService!  

```
import { getConfigService } from '../config';
import { getAuthenticatedHttpClient } from '../auth';
import { getLoggingService } from '../logging';


configureAnalytics(SegmentAnalyticsService, {
  configService: getConfigService(),
  httpClient: getAuthenticatedHttpClient()
  loggingService: getLoggingService(),
});
```

Something like that.

src/analytics/README.rst
```diff
@@ -15,19 +15,20 @@ To install frontend-analytics into your project::
 
 Add the following configuration code to your app::
 
-    import { configureAnalytics, initializeSegment } from '@edx/analytics';
+    import { configure as configureAnalytics, SegmentAnalyticsService } from '@edx/analytics';
```
davidjoy: The `@edx/analytics`, `@edx/frontend-logging`, etc, packages should all be updated here, presumably?

src/analytics/README.rst
abutterworth: 👍

src/analytics/README.rst
abutterworth: 👍

PR #13, opened 2019-11-22
fix: publish dist directory

davidjoy:


.travis.yml
```diff
@@ -3,8 +3,9 @@ node_js: 12
 script:
   - npm run lint
   - npm run test
+  - npm run build
 after_success:
-  - npx semantic-release
+  - npx -p semantic-release -p @semantic-release/git -p @semantic-release/npm -p @semantic-release/commit-analyzer -p @semantic-release/release-notes-generator semantic-release
```
davidjoy: Without explicitly installing all these packages (that's what -p does) first, the "plugins" config above causes semantic-release to fail saying it can't find the packages.  ... This is odd because they're all supposedly shipped _with_ semantic-release.

PR #18, opened 2019-12-02
fix: identify the user after analytics initialization

abutterworth:


src/initialize.js
```diff
@@ -42,14 +42,23 @@ export async function auth(requireUser, hydrateUser) {
...
+
 function applyOverrideHandlers(overrides) {
   const noOp = async () => {};
   return {
     pubSub: noOp,
     config: noOp,
     logging: noOp,
     auth,
-    analytics: noOp,
+    analytics,
```
davidjoy: I'm a bit surprised this didn't break a test.

PR #30, opened 2019-12-05
fix: make axios error messages clearer

abutterworth:
Add request, response, or config to message

### DIFF
--- a/src/auth/utils.js
+++ b/src/auth/utils.js
@@ -69,7 +69,7 @@ const processAxiosError = (axiosErrorObject) => {
       httpErrorRequestUrl,
       httpErrorRequestMethod,
     };
-    error.message = `HTTP Client Error: ${status} ${httpErrorRequestUrl} ${httpErrorResponseData}`;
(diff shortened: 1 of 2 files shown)


src/auth/utils.js
```diff
@@ -78,7 +78,7 @@ const processAxiosError = (axiosErrorObject) => {
       httpErrorRequestUrl,
       httpErrorRequestMethod,
     };
-    error.message = `HTTP Client Error: ${error.message} ${httpErrorRequestMethod} ${httpErrorRequestUrl}`;
+    error.message = `Axios Error (Request): ${error.message} ${httpErrorRequestMethod} ${httpErrorRequestUrl}`;
```
robrap: Add `(possible local connectivity issue)`?
Just wondering how to keep lots of engineers who aren't often in NewRelic from wasting time looking into errors that they shouldn't be.  Note: Doug was one of the first to spend a good chunk of time on this before realizing it wasn't anything we could do anything about, and I don't want others to have to go through the same.

**Alternatives:**
* Leave it and do nothing.  
* Document it somewhere and train people on this?

PR #32, opened 2019-12-05
fix: avoid sending empty translation input files to transifex

abutterworth:
This helps prevent resetting all translations to unreviewed on transifex if something goes wrong in the i18n extract translations process


abutterworth:
This is in place of. I spoke with David and we agreed it would be better to have an explicit check that the object is empty rather than assuming 2 characters in the file means the object is empty.

PR #34, opened 2019-12-06
Generating API Documentation

davidjoy:
We now use a package called "documentation" to generate JSDoc-annotated API documentation as a Markdown file.  RST is not an option for this, and I don't think it's worth translating it as it loses some fidelity around links.


.gitignore
```diff
@@ -4,3 +4,4 @@
 coverage
 dist
 node_modules
+/docs/_API-body.md
```
davidjoy: This is an intermediary file that shouldn't be checked-in.  It can be left in the filesystem when using the `docs-watch` package.json script, though.

Makefile
```diff
@@ -1,3 +1,8 @@
+# Documentation CLI usage: https://github.com/documentationjs/documentation/blob/master/docs/USAGE.md
+
+doc_command = ./node_modules/.bin/documentation build src -g -c ./docs/documentation.config.yml -f md -o ./docs/_API-body.md --sort-order alpha
```
abutterworth: wow

docs/API.md
```diff
@@ -0,0 +1,1360 @@
...
+### Table of Contents
+
+-   [Platform APIs][1]
+    -   [Initialization][2]
+        -   [analytics][3]
+        -   [auth][4]
+        -   [history][5]
+        -   [initError][6]
+        -   [initialize][7]
+    -   [Config][8]
```
abutterworth: Config, PubSub, and Init are all top level exports right? Does this doc reflect that?

PR #64, opened 2020-02-21
fix: define logout URLs correctly (not as login URL dups)

mikix:
Am I crazy?


davidjoy:
Ack!  Please merge. 👍

PR #65, opened 2020-02-25
feat: More flexible page routes

### LINKED ISSUE #63: Support render prop for PageRoute
PageRoute currently requires a `component` prop. But if you want to inspect a match param or whatever, the docs say you should instead use a `render` prop to avoid tearing down your component every render.

For example, from frontend-app-publisher:
```
        <Route
          path="/courses/:uuid"
          exact
          render={({ match }) => (
            <EditCourse uuid={match.params.uuid} />
          )}
        />
```

(Separately, PageRoute itself should use a `render` prop for its component wrapper method that sends page events, yah?)

davidjoy:
- Added redirectUrl prop to AuthenticatedPageRoute to enable redirection to pages other than login.
- Refactored PageRoute to allow all react-router-dom Route props
- Added getLoginRedirectUrl and getLogoutRedirectUrl for direct access to login/logout URL creation.

This fixes #63

### DIFF
--- a/src/react/AuthenticatedPageRoute.jsx
+++ b/src/react/AuthenticatedPageRoute.jsx
@@ -1,9 +1,10 @@
 import React, { useContext } from 'react';
-import { Route } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { useRouteMatch } from 'react-router-dom';
 
 import AppContext from './AppContext';
-import LoginRedirect from './LoginRedirect';
 import PageRoute from './PageRoute';
+import { getLoginRedirectUrl } from '../auth';
 
 /**
  * A react-router route that redirects to the login page when the route becomes active and the user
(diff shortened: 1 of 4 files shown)


src/react/AuthenticatedPageRoute.jsx
```diff
@@ -1,32 +1,60 @@
...
+ * @param {string} props.redirectUrl The URL anonymous users should be redirected to, rather than
+ * viewing the route's contents.
  */
-export default function AuthenticatedPageRoute(props) {
+export default function AuthenticatedPageRoute({ redirectUrl, ...props }) {
   const { authenticatedUser } = useContext(AppContext);
 
+  const match = useRouteMatch({
+    // eslint-disable-next-line react/prop-types
+    path: props.path,
```
adamstankiewicz: Any particular reason we're not adding these props to PropTypes?

src/react/AuthenticatedPageRoute.jsx
```diff
@@ -1,32 +1,60 @@
...
+    // eslint-disable-next-line react/prop-types
+    exact: props.exact,
+    // eslint-disable-next-line react/prop-types
+    strict: props.strict,
+    // eslint-disable-next-line react/prop-types
+    sensitive: props.sensitive,
+  });
+
   // We need to let Route "render" the redirect - if we did it here in AuthenticatedRoute, it'd
```
adamstankiewicz: Is this comment still relevant? Seems like it wouldn't be now that we're no longer rendering the `Route` component here.

src/react/AuthenticatedPageRoute.jsx
davidjoy: It felt a bit weird to reproduce the react-router props when they're kinda supposed to just be a passthrough... but I could be convinced? ;)

src/react/AuthenticatedPageRoute.jsx
davidjoy: Whoop!  Good catch.  Some of these comments got a bit stale.

davidjoy:
@mikix -

`useRouteMatch` is a hook exposed by react-router that lets us use the router's `match` object outside the context of a Route. This was great for this use case, as the only other alternative to know if the page event should be sent was to use render props.  This lets us say "is it a match?" and _then_ decide what we want to render/call, rather than trying to insert a call as a child of the Route itself.  Made the whole thing simpler.

`useEffect` is a memoized side-effect handler.  The second argument, an array, are basically the arguments you want to memoize on.  So this:

```
useEffect(() => {
    if (match) {
      sendPageEvent();
    }
  }, [match]);
```

Is saying, "run this function whenever `match` changes".  Inside the function we send the page event if match isn't null.  

Something like that!

src/react/AuthenticatedPageRoute.jsx
davidjoy: I didn't want to presume required/defaults, too.

PR #66, opened 2020-02-27
export PageRoute and OptionalReduxProvider

adamstankiewicz:
Two of the React components in @edx/frontend-platform are not exported and currently unusable by a consuming application (the import comes back as undefined).

This PR ensures `PageRoute` and `OptionalReduxProvider` are exported.

### DIFF
--- a/src/react/index.js
+++ b/src/react/index.js
@@ -12,4 +12,6 @@ export { default as AuthenticatedPageRoute } from './AuthenticatedPageRoute';
 export { default as ErrorBoundary } from './ErrorBoundary';
 export { default as ErrorPage } from './ErrorPage';
 export { default as LoginRedirect } from './LoginRedirect';
+export { default as OptionalReduxProvider } from './OptionalReduxProvider';
+export { default as PageRoute } from './PageRoute';
 export { useAppEvent } from './hooks';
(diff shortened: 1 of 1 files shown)


davidjoy:
Hmm, I'm not sure `OptionalReduxProvider` should be exported... it's used by AppProvider internally.  Do you have a use case for needing it?  I'd expect app provider to be flexible enough that one wouldn't be needed... 🤔 

And that said, if someone actually knows they want a redux provider and can't use the one in AppProvider somehow, they probably wouldn't need it to be optional?

adamstankiewicz:
> Hmm, I'm not sure `OptionalReduxProvider` should be exported... it's used by AppProvider internally. Do you have a use case for needing it? I'd expect app provider to be flexible enough that one wouldn't be needed... 🤔
> 
> And that said, if someone actually knows they want a redux provider and can't use the one in AppProvider somehow, they probably wouldn't need it to be optional?

Yeah, I wasn't sure if `OptionalReduxProvider` should be exported but figured I'd put it in the PR for feedback on that. Happy to remove it!

I only noticed these components weren't exported when trying to import `PageRoute`, which I assume we want to expose to consuming applications.

davidjoy:
Absolutely - not exporting `PageRoute` was a total oversight. :)

adamstankiewicz:
Cool. I removed the `OptionalReduxProvider` export, and kept the `PageRoute` export. Feel free to take another look.

PR #69, opened 2020-03-13
Adding a custom message to the ErrorPage

staubina:


src/react/ErrorPage.jsx
```diff
@@ -28,12 +28,17 @@ class ErrorPage extends Component {
                 description="error message when an unexpected error occurs"
               />
             </p>
+            {message !== null ? (
```
abutterworth: What do you think about replacing "An unexpected error occurred. Please click the button below to refresh the page." if a message is supplied?

src/react/ErrorPage.jsx
```diff
@@ -43,7 +48,21 @@ class ErrorPage extends Component {
                 />
               }
             />
+            <Button
```
abutterworth: Maybe it would be better to add a prop to be able to add buttons? Do you think Go Back should always be there?

src/react/ErrorPage.jsx
davidjoy: This check may not work - the prop will be `undefined` and not `null` if it hasn't been supplied.  If we add 
```
ErrorPage.defaultProps = {
  message: null,
};
```
At the end of the file, it'll then work.  Also, you may need to add `ErrorPage.propTypes` to pass the linter, too.

src/react/ErrorPage.jsx
```diff
@@ -43,7 +48,21 @@ class ErrorPage extends Component {
...
+              buttonType="primary"
+              onClick={() => {
+                history.back()
+              }}
+              label={
+                <FormattedMessage
```
davidjoy: Just a reminder for us all that frontend-platform doesn't yet have its own translations. :) ¯\_(ツ)_/¯

src/react/ErrorPage.jsx
```diff
@@ -28,12 +28,17 @@ class ErrorPage extends Component {
                 description="error message when an unexpected error occurs"
               />
             </p>
+            {message !== null ? (
+              <div>
```
wittjeff: Could you please give this line role="note" (if you think the message is something that can be ignored) or role="alert" (if you think the message is something everyone would want to read (sounds more likely)?

wittjeff: That is unless it's coming in with <formatted message> already....

src/react/ErrorPage.jsx
staubina: Removed the button for now.

src/react/ErrorPage.jsx
staubina: done

src/react/ErrorPage.jsx
adamstankiewicz: Why not have this conditional be:

```jsx
{message && (
  <div role="alert">
    <p>{message}</p>
  </div>
)}
```

With this, the `div` would only be rendered if `message` is truthy (handling `null`, `undefined`, and empty string), and we wouldn't need to return the empty React fragment in the ternary conditional.

src/react/ErrorPage.jsx
```diff
@@ -28,12 +30,17 @@ class ErrorPage extends Component {
...
+                <p>{message}</p>
+              </div>
+            ) : (<React.Fragment key="nothing"></React.Fragment>)}
           </div>
         </div>
         <div className="row">
           <div className="col">
             <Button
-              buttonType="primary"
+              buttonType="primary mr-2"
```
adamstankiewicz: Is this right margin intentional? If so, you might want to move it the `className` prop. The `buttonType` prop technically should only the Bootstrap button types, with additional class names coming from the `className` prop.

staubina: meant to remove this.  It was some ui testing when I added another button. I will remove.

PR #77, opened 2020-04-07
fix: Changing the name to match what is set in other places

staubina:


src/config.js
```diff
@@ -37,7 +37,7 @@ let config = {
   BASE_URL: process.env.BASE_URL,
   CREDENTIALS_BASE_URL: process.env.CREDENTIALS_BASE_URL,
   CSRF_TOKEN_API_PATH: process.env.CSRF_TOKEN_API_PATH,
-  DISCOVERY_BASE_URL: process.env.DISCOVERY_BASE_URL,
+  DISCOVERY_API_BASE_URL: process.env.DISCOVERY_API_BASE_URL,
```
davidjoy, line 40: There's a JSDoc comment further down in this file that needs to be updated too, I think.

staubina:
@davidjoy Fixed

PR #81, opened 2020-04-21
fix: ceasing to throw errors on undefined config

### LINKED ISSUE #70: ensureDefinedConfig error during initialization is unloggable
```
initialize({
  handlers: {
    config: () => {
      mergeConfig({
        SUPPORT_URL: process.env.SUPPORT_URL,
        COACHING_ENABLED: process.env.COACHING_ENABLED,
      }, 'App loadConfig override handler');
    },
  },
});
```

If someone uses `mergeConfig` - which delegates to `ensureDefinedConfig` - in the `config` handler of their `initialize` call, we have an issue where an `undefined` key causes `ensureDefinedConfig` to throw an Error... which we try to log using `logError`.  Unfortunately, logging hasn't been configured yet, so we get:

```
Uncaught (in promise) TypeError: Cannot read property 'logError' of null
```

Instead of anything useful.

The point of `ensureDefinedConfig` is to catch situations where we forgot to define a key in process.env, so the fact that it explodes is _good_, but we're just effectively swallowing the useful error message.  We should fix that and log to the console if logging isn't configured yet, maybe.

davidjoy:
Also adding PUBLISHER_BASE_URL to config.  Wanting to add this config made me realize that it was going to cause applications to start throwing errors on startup, since none of them would have PUBLISHER_BASE_URL defined yet.  That seems like an unaccpetable, dangerous outcome of changing the config variables which one wouldn’t expect to be a breaking change.

Fixes #70 by avoiding the problem.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,6 +4,7 @@ BASE_URL=localhost:8080
 CREDENTIALS_BASE_URL=http://localhost:18150
 CSRF_TOKEN_API_PATH=/csrf/api/v1/token
 DISCOVERY_API_BASE_URL=http://localhost:18381
+PUBLISHER_BASE_URL=http://localhost:18400
 ECOMMERCE_BASE_URL=http://localhost:18130
 LANGUAGE_PREFERENCE_COOKIE_NAME=openedx-language-preference
 LMS_BASE_URL=http://localhost:18000
--- a/.env.test
+++ b/.env.test
@@ -4,6 +4,7 @@ BASE_URL=localhost:8080
 CREDENTIALS_BASE_URL=http://localhost:18150
 CSRF_TOKEN_API_PATH=/csrf/api/v1/token
 DISCOVERY_API_BASE_URL=http://localhost:18381
+PUBLISHER_BASE_URL=http://localhost:18400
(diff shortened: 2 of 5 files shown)


davidjoy:
@abutterworth Hmm, if PUBLISHER_BASE_URL is undefined in `.env`, doesn't that mean it'll just be undefined when it comes through?

abutterworth:
Yes, but that means we'll have a runtime error somewhere in the life of the
code, likely upon making some kind of request vs blowing up immediately. It
*might be harder to track down config issues. Just a thought – not sure.

davidjoy:
The new behavior logs warnings to the console, rather than throwing an error.  So it _should_ be pretty obvious that a particular variable is broken if anyone looks at the console at all.  Currently it quietly blows up because it can't log an error to a logging service that doesn't exist yet,(See #70), so this is actually an improvement, I think. :)

I don't feel terribly strongly one way or the other... whacha think, should we do the harder refactor and find a way to make it continue to be a fatal error, but with better messaging?  Maybe... 🤔

abutterworth:
Sounds good to me. Obviously we can adjust if we discover any problems :)

PR #84, opened 2020-05-12
Add service links to config for Studio and Insights

stvstnfrd:
Seems reasonable (and in keeping with the pattern already in use) for this config to be aware of the `*_BASE_URL` of each of the various services in the Open edX ecosystem, no?

See:
- `ECOMMERCE_BASE_URL`
- `LMS_BASE_URL`
- etc...

This came up with and facilitates:
- https://github.com/edx/frontend-app-learning/pull/65


marcotuts:
(Is there a base url for Insights in devstack or whatever it is that is relying on that parameter?)

stvstnfrd:
@marcotuts I don't see the Insights (Analytics Dashboard) in the devstack/docker config, just the pipeline :shrug: 
That's what I left it blank here..

This did prompt me to check the `lms.env.json` config inside the devstack, which _does_ set `"ANALYTICS_DASHBOARD_URL": "http://localhost:18110/courses"`

but Insights isn't running there, at least in my devstack.

I edited the value locally during testing (back on https://github.com/edx/frontend-app-learning/pull/65), but didn't want to commit it if it wasn't actually working for me.

.env.development
```diff
@@ -6,6 +6,7 @@ CSRF_TOKEN_API_PATH=/csrf/api/v1/token
 DISCOVERY_API_BASE_URL=http://localhost:18381
 PUBLISHER_BASE_URL=http://localhost:18400
 ECOMMERCE_BASE_URL=http://localhost:18130
+INSIGHTS_BASE_URL=
```
davidjoy, line 9: I think we may want to put a URL here even if it's a bit wrong... not sure what happens exactly if this is undefined like this.  Same with .env.test below.

src/config.js
```diff
@@ -41,6 +41,7 @@ let config = {
   PUBLISHER_BASE_URL: process.env.PUBLISHER_BASE_URL,
   ECOMMERCE_BASE_URL: process.env.ECOMMERCE_BASE_URL,
   ENVIRONMENT,
+  INSIGHTS_BASE_URL: process.env.INSIGHTS_BASE_URL || null,
```
stvstnfrd, line 44: Okay, it took me longer than it should have, but I finally reconciled the difference between what I wanted to do here and how the existing system is expected to work.

The question now, somewhat philosophical, is where to implement the changes; here in `frontend-platform` or back in `frontend-app-learning`?

My thinking for these two config variables, was that they should be provided to all services, but they not be required to be set. In fact, I considered that a "feature not a bug" for now.
- If you don't set `INSIGHTS_BASE_URL`, you just don't get the Insights link.
- If you don't set `STUDIO_BASE_URL`, you just don't get the Studio link.

So here, these values are preferred, not required.
Their absence toggles _off_ the dependent feature.

Insights, particularly, we want to make this toggle-able, since not all community installations run Insights, due to its complexity :(

So since the `ensureConfig` and `ensureDefinedConfig` helpers only check that the values not be `undefined`, we can keep them "empty" by explicitly coalescing `falsey` values to `null` (or empty string, if we're concerned about null-check issues).

So, do we think the default `getConfig` should be able to provide empty defaults (code as is here with the latest commit and the `|| null`)?

Or do we want to draw a line in the sand and leave it up the implementing app (move these changes to a `mergeConfig` call in `frontend-app-learning`)?

```js
mergeConfig({
  'INSIGHTS_BASE_URL': process.env.INSIGHTS_BASE_URL || null,
  // ...
});
```

.env.development
stvstnfrd: See notes below about intentionally trying to keep this value empty.

davidjoy: Huh, re: running a local branch of frontend-platform with a local app, there are some instructions on the README here: https://github.com/edx/frontend-platform for testing a checkout of frontend-platform with a local app - did those not work?  😬

stvstnfrd:
Closing in favor of a more localized approach in https://github.com/edx/frontend-app-learning/pull/65

PR #91, opened 2020-06-03
feat: Auth module interface creation

davidjoy:
NOTE: This PR technically includes a **backwards incompatible change** to the method signature of the `configure` function in the frontend-platform/auth package.  I've chosen NOT to mark this as a major version bump of the package because I can find no evidence in the project's dependents that any external projects are using this API.  To do so, they would have had to forego using the `initialize` function and instead manually configured the auth service.  If you're doing this and seeing errors that various types of configuration are undefined, please adjust your call to `configure`.  

This PR is best reviewed commit-by-commit.  Please see commit messages and the summaries below for context on each.

1fa8618 - Moving auth service out of index. 

This commit merely moves the guts of the auth service out of 'index.js' and into 'AxiosJwtAuthService.js'.  It then modifies index.js to export the exact same things, but from that file instead.  

b27ac24 - Improving shape of auth configure.

This commit makes the auth service's `configure` method more closely match our other services like logging and analytics, in preparation for turning it into a constructor.  It changes the signature by packaging the config-related variables under `config` in the `options` object.

Note that this is _technically_ a breaking change for anyone using the auth service package directly.  That said, I've looked at the dependents of this project and don't see anywhere that that's being done.  Everyone is creating their auth service through the top level `initialize` function.  **So I'm _not_ going to be marking this PR as a breaking change, but will document this change in the release notes just in case.**  

8941396 - First pass at creating AxiosJwtAuthService. 

This is the big switch - adds an `interface.js` that gets exported, and packages the majority of the axios-based implementation into AxiosJwtAuthService as a class.  The implementation still relies on singletons and such for getCsrfToken and getJwtToken, so it's not complete, but much closer. 

66a0c0b - Folding addAuthenticationToHttpClient in as a static method. 

In an effort to make the service more class-like, this pulls `addAuthenticationToHttpClient` into the class as a static method.  In a later commit, I end up turning it into an instance method instead as it needs to make use of too many instance variables for it to reasonably be considered static anymore.  

50b1a94 - Converting getJwtToken into a class. 

This is one of the more hairy commits, as this code is a bit brain-bending to begin with.  getJwtToken.js creates `httpClient` and `cookies` in the top level file scope, meaning that they're created once and re-used no matter what we did in `AxiosJwtAuthService` to package everything into an instantiatable class.  So this PR takes the approach of turning getJwtToken into an `AxiosJwtTokenService` which then has its httpClient and cookies objects as instance variables.  This has the side effect of meaning we need to pass around less data between the methods in this file, which is nice, since we can just get them from `this` instead.

bf68bdc - All auth singletons gone. 

This is where we finally solve the loggingService issues, though the title of the commit is a misnomer - we still have getCsrfToken to deal with.

We have various other files in the implementation that all import `getLoggingService` from what's now AxiosJwtAuthService - this keeps us from properly making that service an instance at best, and at worst puts us at risk for circular dependency issues.

This solves the problem by passing a reference to the loggingService around so it can be used - it doesn't feel quite ideal to pass it into `logFrontendAuthError`, but that seemed a minor compromise to simplify the rest.  

This also converts the processAxiosRequestErrorInterceptor interceptor to use a 'create' style factory function that allows us to save loggingService in its lexical scope like our other interceptor creators.  

c02d7a0 - Normalizing config variable names in auth 

This is the other half of the breaking change - rather than rename our config variables to camelCase from SCREAMING_SNAKE_CASE, we just pass the whole config object in.  This means that other implementations of the auth service could expect different variables, which is nice.  

**Again, I can see no dependents of this project that rely on this method signature, so I'm _not_ marking this as a breaking change to avoid a whole lot of manual updating of apps.**

f8edbe6 - NOW the auth service is a proper set of classes. 

This cleans up getCsrfToken in the same way we cleaned up getJwtToken - it's now an instantiatable class called AxiosCsrfTokenService.

51a48dd - Clean up and fixing broken docs. 

This commit fixes the broken docs build.


src/auth/AxiosJwtAuthService.js
```diff
@@ -0,0 +1,303 @@
...
+import { logFrontendAuthError } from './utils';
+import { camelCaseObject, ensureDefinedConfig } from '../utils';
+import createJwtTokenProviderInterceptor from './interceptors/createJwtTokenProviderInterceptor';
+import createCsrfTokenProviderInterceptor from './interceptors/createCsrfTokenProviderInterceptor';
+import createProcessAxiosRequestErrorInterceptor from './interceptors/createProcessAxiosRequestErrorInterceptor';
+import AxiosJwtTokenService from './AxiosJwtTokenService';
+import AxiosCsrfTokenService from './AxiosCsrfTokenService';
+
+const optionsPropTypes = {
+  config: PropTypes.shape({
```
davidjoy, line 12: Note, we've changed the options signature here.  This is _technically_ a breaking change, but I'm consciously not marking it as such because no dependents of this project rely on `configure` directly - they all have it done for them via `initialize`.

src/auth/AxiosJwtAuthService.js
```diff
@@ -0,0 +1,303 @@
...
+ * Adds authentication defaults and interceptors to an HTTP client instance.
+ *
+ * @param {HttpClient} newHttpClient
+ * @param {Object} config
+ * @param {string} [config.REFRESH_ACCESS_TOKEN_ENDPOINT]
+ * @param {string} [config.ACCESS_TOKEN_COOKIE_NAME]
+ * @param {string} [config.CSRF_TOKEN_API_PATH]
+ * @returns {HttpClient} A configured Axios HTTP client.
+ */
+  addAuthenticationToHttpClient(newHttpClient) {
```
davidjoy, line 252: This function takes the HTTP client to apply authentication to, but otherwise uses the configuration of this AxiosJwtAuthService instance to do it.  That feels like a reasonable method signature to me - it used to be a static function, but that was back when everything was static/singletons.

src/auth/AxiosJwtAuthService.js
```diff
@@ -0,0 +1,303 @@
...
+      csrfTokenService: this.csrfTokenService,
+      CSRF_TOKEN_API_PATH: this.config.CSRF_TOKEN_API_PATH,
+      shouldSkip: (axiosRequestConfig) => {
+        const { method, isCsrfExempt } = axiosRequestConfig;
+        const CSRF_PROTECTED_METHODS = ['post', 'put', 'patch', 'delete'];
+        return isCsrfExempt || !CSRF_PROTECTED_METHODS.includes(method);
+      },
+    });
+
+    const processAxiosRequestErrorInterceptor = createProcessAxiosRequestErrorInterceptor({
```
davidjoy, line 281: This is a change - this interceptor didn't have a `create*` function prior to this.

src/auth/AxiosJwtAuthService.test.jsx
```diff
@@ -96,19 +81,18 @@ const mockApiEndpointPath = `${process.env.BASE_URL}/api/v1/test`;
 window.location.assign = jest.fn();
 const mockCookies = new Cookies();
 
-// This sets the mock adapter on the default instance
```
davidjoy, line 99: We re-set all these variables on `beforeEach` now, since they need to be re-created in sync with each other for every test.  If I left them here, they'd fall out of sync when we re-configure/instantiate the `AxiosJwtAuthService` for each test.

src/auth/AxiosJwtAuthService.test.jsx
```diff
@@ -2,39 +2,24 @@
 import axios from 'axios';
```
davidjoy, line 2: @abutterworth - Bang up job on these tests.  It's a testament to how well written they are that I barely had to change them to have them work with the new code shape.

src/auth/interface.js
```diff
@@ -0,0 +1,253 @@
...
+export function getAuthenticatedUser() {
+  return service.getAuthenticatedUser();
+}
+
+/**
+ *
+ */
+export function setAuthenticatedUser(authUser) {
+  service.setAuthenticatedUser(authUser);
+  publish(AUTHENTICATED_USER_CHANGED);
```
davidjoy, line 173: Note how this `publish` call is part of the interface.  This is a choice we made that the interface can provide extra functionality like this, which allows the implementation to be agnostic and more unaware of larger concerns of the frontend-platform as a whole.

src/auth/utils.js
```diff
@@ -31,10 +29,10 @@ const getUrlParts = (url) => {
   }
 };
 
-const logFrontendAuthError = (error) => {
+const logFrontendAuthError = (loggingService, error) => {
```
davidjoy, line 32: As mentioned elsewhere, this is a small compromise for simplicity - we pass the loggingService through here so that this function can stay as a stand-alone helper.  Otherwise it'd need to be attached to an instance so it could do `this.loggingService`, and we also need it in multiple classes.

src/initialize.js
```diff
@@ -247,6 +247,7 @@ function applyOverrideHandlers(overrides) {
 export async function initialize({
   loggingService = NewRelicLoggingService,
   analyticsService = SegmentAnalyticsService,
+  authService = AxiosJwtAuthService,
```
davidjoy, line 250: Hey, you can configure an alternate auth service now!

binodpant:
General question: Would you mind adding some test notes on how best we can verify/test this change locally? Are there any specific test cases to consider?

binodpant:
Will this uage still be supported? I see it in an app I am working on 

```
import {
  initialize,
  APP_INIT_ERROR,
  APP_READY,
  subscribe,
} from '@edx/frontend-platform';
```

binodpant:
Let's also update the main README paragraph that says the following?

> NOTE: As of this writing, both authentication and i18n are not configurable. The initialize() function does not allow applications to supply alternate auth/i18n implementations; this is because the interface and implementations for auth and i18n have not yet been separated and modularized.

binodpant:
@davidjoy I guess it would be worth posting about this particular change to the public api method somewhere before this is sent to prod? Just in case anyone was hacking around in their dev and using it! unlikely I presume but possible?

PR #92, opened 2020-06-08
feat: Adding a MockAuthService, improving docs

davidjoy:
The mock auth service uses axios-mock-adapter in its implementation to allow mocking of any http client calls the auth service might make on behalf of an application.

This PR also fleshes out the docs for the auth service to try to add some detail around the differences between the interface, AxiosJwtAuthService, and new MockAuthService implementations.

### DIFF
--- a/src/auth/MockAuthService.js
+++ b/src/auth/MockAuthService.js
@@ -1,0 +1,275 @@
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+import PropTypes from 'prop-types';
+import { ensureDefinedConfig } from '../utils';
+
+const userPropTypes = PropTypes.shape({
+  userId: PropTypes.string.isRequired,
+  username: PropTypes.string.isRequired,
+  roles: PropTypes.arrayOf(PropTypes.string),
+  administrator: PropTypes.boolean,
+});
+
+const optionsPropTypes = {
+  config: PropTypes.shape({
+    BASE_URL: PropTypes.string.isRequired,
+    LMS_BASE_URL: PropTypes.string.isRequired,
+    LOGIN_URL: PropTypes.string.isRequired,
+    LOGOUT_URL: PropTypes.string.isRequired,
+    REFRESH_ACCESS_TOKEN_ENDPOINT: PropTypes.string.isRequired,
+    ACCESS_TOKEN_COOKIE_NAME: PropTypes.string.isRequired,
+    CSRF_TOKEN_API_PATH: PropTypes.string.isRequired,
+  }).isRequired,
+  loggingService: PropTypes.shape({
+    logError: PropTypes.func.isRequired,
+    logInfo: PropTypes.func.isRequired,
+  }).isRequired,
+  // The absence of authenticatedUser means the user is anonymous.
+  authenticatedUser: userPropTypes,
+  // Must be at least a valid user, but may have other fields.
+  hydratedAuthenticatedUser: userPropTypes,
+};
+
+/**
+ * The MockAuthService class uses axios-mock-adapter to wrap it's HttpClients so that they can be
+ * mocked for development and testing.  In an application, it could be used to temporarily override
+ * the "real" auth service:
+ *
+ * ```
+ * import {
+ *   initialize,
+ *   APP_AUTH_INITIALIZED,
+ *   subscribe,
+ *   mergeConfig,
+ * } from '@edx/frontend-platform';
+ * import { MockAuthService } from '@edx/frontend-platform/auth';
+ *
+ * initialize({
+ *   handlers: {
+ *     config: () => {
+ *       mergeConfig({
+ *         authenticatedUser: {
+ *           userId: 'abc123',
+ *           username: 'Mock User',
+ *           roles: [],
+ *           administrator: false,
+ *         },
+ *         hydratedAuthenticatedUser: {
+ *           // Additional user props expected to be returned from a user accounts API, providing
+ *           // additional user details beyond those present in the JWT.  This data is added into
+ *           // the user object if and when hydrateAuthenticatedUser is called.  See documentation
+ *           // for `hydrateAuthenticatedUser` for more details.
+ *         }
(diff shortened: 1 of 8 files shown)


src/auth/MockAuthService.js
```diff
@@ -0,0 +1,273 @@
...
+ * ```
+ *
+ * NOTE: The login/logout methods related to redirecting currently maintain their real behaviors.  A
+ * subsequent update to this mock service could allow them to be configured/mocked via the
+ * constructor's config options.
+ *
+ * @implements {AuthService}
+ * @memberof module:Auth
+ */
+class MockAuthService {
```
binodpant, line 103: SUGGEST/DISCUSS: NONBLOCKING: Just had a thought. Would we be interested in this mock client not being used in production code ever since that would bypass auth? I guess right now one would still add a dependency on auth, as opposed to a devDependency. And can also use this in prod code I think without any restrictions. Outside the scope of this PR but would it be worth having a separate mock adapter dependency that can be pulled in via devDependency instead? Wonder what you think about that. That lib would house this plus any other needed auth utilities (if they exist) that are for use in dev rather than for shipping code

src/auth/MockAuthService.js
```diff
@@ -0,0 +1,273 @@
...
+   * Gets the authenticated HTTP client singleton which is an axios instance.
+   *
+   * @returns {HttpClient} Singleton. A configured axios http client
+   */
+  getAuthenticatedHttpClient() {
+    return this.authenticatedHttpClient;
+  }
+
+  /**
+   * Gets the unauthenticated HTTP lient singleton which is an axios instance.
```
binodpant, line 147: typo `lient` -> `client`

src/auth/MockAuthService.js
```diff
@@ -0,0 +1,273 @@
...
+import { ensureDefinedConfig } from '../utils';
+
+const userPropTypes = PropTypes.shape({
+  userId: PropTypes.string.isRequired,
+  username: PropTypes.string.isRequired,
+  roles: PropTypes.arrayOf(PropTypes.string),
+  administrator: PropTypes.boolean,
+});
+
+const optionsPropTypes = {
```
binodpant, line 13: Can we somehow leverge the optionsShape proptypes defined in https://github.com/edx/frontend-platform/blob/cf35b988872727f8096a95a38d554b9ec739c56d/src/auth/interface.js#L57 here? Are these the same?

src/auth/interface.js
```diff
@@ -117,78 +120,125 @@ export function resetAuthService() {
...
+ * existing authenticatedUser object, available via getAuthenticatedUser().
+ *
+ * ```
+ *  console.log(authenticatedUser); // Will be sparse and only contain basic information.
+ *  await hydrateAuthenticatedUser()
+ *  const authenticatedUser = getAuthenticatedUser();
+ *  console.log(authenticatedUser); // Will contain additional user information
+ * ```
  *
+ * @returns {Promise<null>}
```
binodpant, line 241: not sure if hydrate is expected to fail and if a boolean response instead of null would be better? if that has not come up we can ignore this note :)

binodpant:
Generally looking great to me! The mock service seems highly useful indeed and comforms to the interface.js service shape so should be very handy. Do we want any notes in the README or any other place in the api docs for this? Perhaps even a link to the jsdoc line of this file from readme would suffice for easier locatability of this feature?

src/auth/MockAuthService.js
davidjoy: I don't think this would nor should ever be used in production - i.e., it has no place where a user may actually see it.  I can see it being used temporarily during early development when backend APIs aren't yet available.  But even there, that's not a very strong use case.  The primary use case is for writing tests of code that uses frontend-platform/auth to make HTTP requests.  

Any consuming app that would use this would presumably already have frontend-platform as a non-dev dependency already.  It's arguable that this class could be split out from frontend-platform itself, but I'd advocate against putting it in its own repository, as we already have too much repo overhead as is.

Also, because of webpack tree shaking, apps won't incur any overhead for this class being in frontend-platform unless they're actually using it.

src/auth/MockAuthService.js
davidjoy: Darn, I thought I fixed that. ;)

src/auth/interface.js
davidjoy: `hydrateAuthenticatedUser` doesn't actually resolve its promise with any data; any errors will get thrown/handled as part of the promise itself.  So `null` is pretty much correct, I think. :)

davidjoy:
The `MockAuthService` class will be linked into the frontend-platform documentation as an implementation of the `AuthService` interface, kinda like this: https://edx.github.io/frontend-platform/module-Auth.AxiosJwtAuthService.html

Including it in the README might be a bit too low-level for that doc, IMHO, which tries to give an overview of the whole platform.  I'm also likely to write an ADR on using it for testing redux/thunk/api code, for what it's worth.

src/auth/MockAuthService.js
davidjoy: They're a little different - the mock one has some additional fields you can pass in to mock some user data.  I think we could try to merge them, but I'm not 💯 it's worth tying the two files together... 🤔

src/auth/MockAuthService.js
binodpant: got it thanks!

src/auth/MockAuthService.js
binodpant: Hmm okay I was just wondering about what happens in we change the interface options shape in interface.js, presuming the mock always needs to conform the the interface in interface.js. If the mock is an extension of that interface I guess a kind of merge would make sense. But if that is too much work right now, I think we can farm it off as a separate enhancement maybe...

davidjoy:
I consider this PR as groundwork for more concrete guidance around testing, so yeah, want to make sure people can find this. :) I think it's going to be part of a different task for me, though.

PR #94, opened 2020-06-18
fix: MockAuthService returns normal axios instances

davidjoy:
MockAuthService misused axios-mock-adapter’s MockAdapter class.  It is not a drop-in replacement for an axios instance, but a companion object that, when configured, will intercept requests to the axios instance it was instantiated with.  You cannot call .get() or .post() on the mock itself - you need to call those on the original instance.  Therefore MockAuthService needs to return the real deal, and instantiators are responsible for using axios-mock-adapter if they want to.

This means the class’s primary use is to mock out the authenticated user loading.

### DIFF
--- a/src/auth/MockAuthService.js
+++ b/src/auth/MockAuthService.js
@@ -1,5 +1,4 @@
 import axios from 'axios';
-import MockAdapter from 'axios-mock-adapter';
 import PropTypes from 'prop-types';
 import { ensureDefinedConfig } from '../utils';
 
@@ -31,9 +30,10 @@ const optionsPropTypes = {
 };
 
 /**
- * The MockAuthService class uses axios-mock-adapter to wrap it's HttpClients so that they can be
- * mocked for development and testing.  In an application, it could be used to temporarily override
- * the "real" auth service:
+ * The MockAuthService class mocks authenticated user-fetching logic and allows for manually
+ * setting user data.  It is compatible with axios-mock-adapter to wrap its HttpClients so that
+ * they can be mocked for development and testing.  In an application, it could be used to
+ * temporarily override the "real" auth service:
  *
  * ```
  * import {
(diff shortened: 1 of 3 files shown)


src/auth/MockAuthService.js
```diff
@@ -71,26 +72,36 @@ const optionsPropTypes = {
...
+ *     username: 'Mock User',
+ *     roles: [],
+ *     administrator: false,
+ *   },
+ * });
  * configure(MockAuthService, { config: getConfig(), loggingService: mockLoggingService });
- * const mockAuthenticatedHttpClient = getAuthenticatedHttpClient();
- *
- * // Mock calls for your tests.  This configuration can be done in any sort of test setup.
+ *   const mockAdapter = new MockAdapter(getAuthenticatedHttpClient());
```
adamstankiewicz, line 102: tiny nit: looks like these lines have some extra spaces/indent from the left?

src/auth/MockAuthService.js
```diff
@@ -71,26 +72,36 @@ const optionsPropTypes = {
  *   // This variable is now a MockAdapter from axios-mock-adapter, allowing onGet, onPost, etc.
  *   // mocking.  This handler will be called prior to any further initialization.  See the
  *   // "Application Initialization" phases in the README for call order details.
- *   const mockAuthenticatedHttpClient = getAuthenticatedHttpClient();
+ *   const mockAdapter = new MockAdapter(getAuthenticatedHttpClient());
  *   // Mock calls here.
- * });
+ *   mockAdapter.onGet(...);
+  * });
```
adamstankiewicz, line 78: tiny nit: extra space/indent?

binodpant:
PR itself looks good to me. Just curious here is how we currently mock out auth in our tests for frontend, post this merges, should we re-adapt our mocking?

```
// TODO: Need to confirm if this is the best way to mock auth.
jest.mock('@edx/frontend-platform/auth');
getAuthenticatedUser.mockReturnValue({ username: 'test-username' });

```

right now it looks like we just use regular jest mocking of the entire /auth module but maybe this new approach will help us be cleaner and more flexible?

davidjoy:
I think either is reasonable - it depends on the level of integration needed.  If you're testing a component that _only_ calls `getAuthenticatedUser`, then mocking out the module is certainly easiest.  If, on the other hand, you're working with a component that makes some requests and uses other functions from `@edx/frontend-platform/auth`, it may make more sense to use `MockAuthService`.

PR #117, opened 2020-10-10
fix(auth): ensure app is refreshed after user hydration is complete

adamstankiewicz:
There is a bug since, I believe, v1.4.0 where if the user's account should be hydrated (i.e., more user metadata fetched from an API), the app does not properly refresh once the hydration API request resolved.

This behavior can be seen in the `example` application within this repository.

This PR ensures the app is refreshed such that when the hydration API request resolves, the `authenticatedUser` has the additional user metadata as expected. It still does not block on rendering the app before the hydration API request resolves.

**Testing Instructions**
1. Checkout this branch locally: `astankiewicz/fix-user-hydration-reload`.
2. `npm install`, if necessary.
3. Start the example web app: `cd example && npm start`.
4. Observe the authenticated user's name appears after the hydration request is resolved.

### DIFF
--- a/src/auth/MockAuthService.js
+++ b/src/auth/MockAuthService.js
@@ -225,7 +225,6 @@ class MockAuthService {
    * Sets the authenticated user to the provided value.
    *
    * @param {UserData} authUser
-   * @emits AUTHENTICATED_USER_CHANGED
    */
   setAuthenticatedUser(authUser) {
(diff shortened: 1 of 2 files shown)


davidjoy (approved):
Doh!  Must have missed that in the refactoring.  Nice catch.

PR #119, opened 2020-10-15
Add axios cache adapter to add front end caching to network requests

kellmarie:
As part of ENT-3347 we want to reduce the number of network requests made on each page when the response data doesn't change between calls. Using axios-cache-adapter with a localForage store we're able to cache the response from a network call and reuse the response data that hasn't changed so each page load doesn't keep making the same request.

Response data is primarily stored in IndexedDB for web browsers that have IndexedDB, [according to this website that's the case for the browsers of 93% of internet users](https://caniuse.com/?search=indexeddb). In the small percentage of browsers that don't support IndexedDB the response falls back to localStorage for saving the data.

The default MaxAge for each request is currently 5 minutes, this MaxAge is able to be [changed on a per request basis](https://www.npmjs.com/package/axios-cache-adapter#override-instance-config-with-per-request-options). The default time of 5 minutes can be discussed further it was just picked as a reasonable MaxAge for the time being.


adamstankiewicz (commented):
One question from a quick initial pass: Is it possible to use a single httpClient rather than defining a cached vs. non-cached one? Ideally, it'd be great for the original httpClient to be cache-ready but only when the `cache` option is provided on the individual requests to opt-in into it.

This would remove the complexity around managing 2 different httpClients in frontend-platform.

Perhaps setting the default `maxAge` option to `0` during `setup()` would prevent caching on all requests, unless you override it on individual requests?

iloveagent57 (approved):
Looks great, thanks for the how-to doc.

src/auth/AxiosJwtAuthService.js
```diff
@@ -59,24 +62,41 @@ class AxiosJwtAuthService {
     this.csrfTokenService = new AxiosCsrfTokenService(this.config.CSRF_TOKEN_API_PATH);
     this.authenticatedHttpClient = this.addAuthenticationToHttpClient(axios.create());
     this.httpClient = axios.create();
+    configureCache().then((cachedAxiosClient) => {
```
davidjoy, line 65: I'm wondering if we should add a try/catch around this call to configureCache... if it fails for some reason, we can still proceed with loading, we just might not be able to cache requests.  I'm thinking that perhaps we set the `cachedAuthenticatedHttpClient` and `cachedHttpClient` to the _non-cached_ versions and log an error to new relic via `logError`.

It seems like it might be a good idea, given that `configureCache` is interacting with a bunch of browser APIs which may or may not exist.

davidjoy (changes requested):
Everything looks great, I think - especially the tests - though I'm thinking we should be a bit defensive around `configureCache`

davidjoy (approved):
🎉

PR #122, opened 2020-10-21
fix: bumping version of frontend-build to latest

davidjoy:
Also opportunistically removing a strange script tag loading ‘index.js’ in the example app, which doesn’t work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "devDependencies": {
     "@commitlint/cli": "8.2.0",
     "@commitlint/config-angular": "8.2.0",
-    "@edx/frontend-build": "^5.3.2",
+    "@edx/frontend-build": "5.3.2",
     "@edx/paragon": "7.1.5",
     "axios-mock-adapter": "1.17.0",
     "babel-polyfill": "6.26.0",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "devDependencies": {
     "@commitlint/cli": "8.2.0",
     "@commitlint/config-angular": "8.2.0",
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
   "devDependencies": {
     "@commitlint/cli": "8.2.0",
     "@commitlint/config-angular": "8.2.0",
-    "@edx/frontend-build": "2.0.6",
+    "@edx/frontend-build": "^5.3.2",
```
adamstankiewicz, line 36: Should we pin this dependency so we're consistently pinning all dependencies?

davidjoy: Fixed!

src/react/AuthenticatedPageRoute.test.jsx
```diff
@@ -45,6 +47,7 @@ describe('AuthenticatedPageRoute', () => {
       </AppContext.Provider>
     );
     history.push('/authenticated');
+    global.location.href = 'http://localhost/authenticated';
```
adamstankiewicz, line 50: nit: I'm curious why it's necessary to update `global.location.href` now when it wasn't needed before. Is it just because we're now doing `delete global.location;` rather than `delete global.location.assign;`, which wouldn't affect `location.href`?

adamstankiewicz (approved):
Looks good! I added one nit/question, but I don't think it is blocking.

src/react/AuthenticatedPageRoute.test.jsx
davidjoy: Yeppers!

PR #128, opened 2020-11-15
add config var for logistration mfe

uzairr:


davidjoy:
Hi there - I think there's a better approach that doesn't involve updating frontend-platform.  You can use the `mergeConfig` function in order to add custom config into the object returned by `getConfig`.  The best place to do this is during the `initialize` function, as shown here:

https://github.com/edx/frontend-app-learning/blob/master/src/index.jsx#L82

We try to only add config variables to frontend-platform that are expected to be used across micro-frontends.  The `mergeConfig` mechanism was added for extra config that a frontend needs.

You can see documentation for `initialize` and `mergeConfig` here:

https://edx.github.io/frontend-platform/module-Initialization.html#~initialize
https://edx.github.io/frontend-platform/module-Config.html#~mergeConfig

The example from frontend-app-learning may be the most complete/easy to follow, though!

Let me know if that doesn't solve your needs without needing this.  If it does, I'd prefer to close this PR.  Thanks!

waheedahmed:
Thank you so much for your input @davidjoy!

I have verified your suggestion and it works fine according to our needs, closing this PR out.

PR #131, opened 2020-11-18
Load Braze from their CDN when the keys and endpoint are defined

cpennington:


davidjoy:
Is this something we intend to do on all/most MFEs?  If this is for a particular MFE, the app initialization process publishes a `APP_ANALYTICS_INITIALIZED` event which allows you to execute code immediately after the analytics service has initialized, which should be effectively equivalent to the code here.  

This could look like this in index.jsx of the app:

```
initialize({
  handlers: {
    config: () => {
      mergeConfig({
        BRAZE_ENDPOINT: process.env.BRAZE_ENDPOINT || null,
        BRAZE_KEY: process.env.BRAZE_KEY || null,
      }, 'Braze Config');
    },
  },
  // ... other initialization options
});

function initializeBraze() {
  // ... function from this PR
}

subscribe(APP_ANALYTICS_INITIALIZED, () => {
  if (getConfig().BRAZE_KEY && getConfig().BRAZE_ENDPOINT) {
     initializeBraze();
    }
});
```

(I didn't test the above, so may contain typos)

davidjoy:
Going to close this PR, as I think it's just been hanging out for a while and may not belong in the library.  

(Best wishes on the new gig, Cale!)

PR #133, opened 2020-11-30
Decision document for the axios cache layer

kellmarie:


docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+The titans team noticed that on a lot of pages the same API calls are repeatedly being made as the user navigates around the website. The data being returned from these requests doesn't change between page loads leading to the same endpoints reloading the same data repeatedly. We decided that we wanted to avoid repeatedly reloading the data and reuse the data that was in the response from the initial load.
```
iloveagent57, line 12: Nit: you don't have to say "the titans team", here - it's potentially confusing to a reader outside of our realm.  Plus, team names change, teams come and go, etc.

iloveagent57, line 12: nit: "repeatedly reloading" is kind of redundant, maybe use the phrase "repeatedly loading unchanged data" to represent the thing we want to avoid doing.

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+We think it's beneficial to cut as much overhead as possible so each page becomes interactive more quickly.
+
+We decided to add a frontend caching layer to the AxiosJwtAuthService by leveraging an already existing npm package: `axios-cache-adapter <https://www.npmjs.com/package/axios-cache-adapter>`_
+
+Having a frontend caching layer would also allow for a stale version of a page to be displayed for a short amount of time in the case of a backend outage.
+
+When Frontend Caching Should Be Used
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The frontend cache should be used to eliminate the need to reload the same data on each page load as a user navigates around. It shouldn't be used to reduce the number of times multiple components on the same page try to make the same network call. Reducing the number of times the same network call is made on a single page should be done by moving the call higher up in the hierarchy so the call only gets made once and all lower level components are able to access the response data from it.
```
iloveagent57, line 23: Nit: "The frontend cache should be used to eliminate the need to reload the same data on each page load as a user navigates around" -> "The frontend cache should be used to eliminate the reloading of unchanged data as a user navigates between different pages in the application."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Requests that are good candidates to use frontend caching are ones that:
+ * The data being returned changes very infrequently independently from the actions of the user on the page
```
iloveagent57, line 26: Nit: "...changes very infrequently, and independently from..."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+   - For example changes made by a background job or an admin making changes
```
iloveagent57, line 28: nit: you don't need "making changes" at the end of this sentence.

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+ * The data being returned isn't specific to the authenticated user
+
+ * The data returned is specific to the authenticated user but front end does not make the request when there isn't an authenticated user
```
iloveagent57, line 32: I'm not sure I understand what this sentence means.  Do you mean that it's a request for user-specific data that _only_ takes place when the user is authenticated?  Meaning a user is making an authenticated request for data only about themself?

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+It should not be used as the only caching strategy that gets used. It is preferable to use a serverside caching strategy in conjunction with frontend caching.
```
iloveagent57, line 34: Nit: "should not be used" is redundant with "that gets used"

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+How To Use Frontend Caching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The frontend caching layer can be used by passing {useCache: true} into the function that gets the httpClient.
```
iloveagent57, line 39: Nit: put some double back-ticks around `{useCache: true}`

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+e.g.
+    service.getAuthenticatedHttpClient({ useCache: true });
```
iloveagent57, line 42: Nit: same thing here, format it as code.

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+By default the response for a GET request will be stored in IndexedDB for 5 minutes and that cached value will get invalidated on any POST, PUT, PATCH, or DELETE request made to the same url. The caching layer also works with the standard cache-control headers: max-age, no-cache, and no-store
+
+Each request can also override the default cache configurations. `The How To document about caching <https://github.com/edx/frontend-platform/blob/master/docs/how_tos/caching.rst>`_ has more detailed examples about how this caching layer can be used.
+
+
+Implementation Details
+~~~~~~~~~~~~~~~~~~~~~~
+
+For both versions of the current AxiosJwtAuthService HTTP clients, the authenticated client and the base unauthenticated clients, there are now two additional versions that use the cache (i.e. CachedAuthenticated and CachedUnauthenticated). Bringing the total number of httpClient instances up to four.
+In order to keep the interface more simple the cached clients don’t have their own getter functions exposing this detail, in order to use the cached http client micro frontends pass in a param for using the cache with the current getter functions.
```
iloveagent57, line 53: Suggest changing this to
"In order to simplify the interface, the cached clients don’t have their own getter functions. In order to use the cached version of these HTTP clients, an application must pass to the existing client getter methods a parameter indicating that cached data is acceptable, for example, using `{useCache: true}`."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+In the case of the cached clients failing to get created the uncached clients will be returned and used instead.
```
iloveagent57, line 55: "In the case of failure during creation of cached client objects, an uncached client object will be returned and used instead."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+The async hydrateAuthenticatedUser function will continue only ever using the uncached authenticated HTTP client so there won't be any concerns about the cache changing expected behavior of authentication.
```
iloveagent57, line 57: Nit: double backticks around `hydrateAuthenticatedUser`

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Only the AxiosJwtAuthService currently has cached http clients.
+
+Decisions
+---------
+
+By using the existing axios-cache-adapter we were able to avoid needing to implement our own caching strategy and forgo having micro frontends need to implement their own front end caching strategies.
+
+The default max-age for cached data was decided to be 5 minutes for now. The decision was based on it being a decent amount of the storage of cached data even for authentication specific requests. The assumption is made that sers won’t be frequently switching to different accounts on their browser frequently enough to have a wide impact and if they do then they will only see cached data for a maximum of 5 minutes. Micro frontends are also able to override the cache time to suit their own needs.
```
iloveagent57, line 66: Nit: delete "for now".  If it changes someday, it changes, and we can update this ADR or write a new one (or both).

iloveagent57, line 66: typo: "sers" -> "users"

iloveagent57, line 66: Couple other points here:
* "The decision was based on it being a decent amount of the storage of cached data even for authentication specific requests." I'm not sure I understand this sentence, could you elaborate a little bit?
* You don't need the first "frequently"
* I think the third sentence could use a comma or two.
* "...override the cache time..." -> "...override the default cache expiry..."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+The default max-age can also be changed at a later time if the current default of 5 minutes is deemed to be inadequate.
+
+Tests written to verify the cache adapter integrates well with the current interceptors make real external API calls to https://httpbin.org/ as per the tests in the axios-cache-adapter source code.
```
iloveagent57, line 69: Suggest changing to: "Integration tests were added to ensure that the cache adapter does not break the current behavior of the existing interceptors. These tests make real HTTP requests to https://httpbin.org/ as suggested by the tests in the `axios-cache-adapter` source code."

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+It was found to be impossible to mock out the exact spot the cache client makes an outside request without also mocking over the cache clients caching functionality. This wasn’t the preferred approach to testing because it makes external network requests but since it is the approach that was used in the axios-cache-adapter tests it was decided to be fine.
+
+Alternatives
+------------
+
+We discussed the pros and cons of having a front-end platform cache implementation for micro front ends to use versus micro front ends implementing their own front end caching strategies utilizing Local/Session storage, ETags, and endpoints utilizing Cache-Control headers.
```
iloveagent57, line 75: Why did we not embrace any of these alternatives?  Or could these be used in conjunction with the implementation we settled on?

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Adoption
+--------
+
+As of this writing, only a few HTTP requests within frontend-app-learner-portal-enterprise have adopted using the cached clients. Usage of the cache clients is on an opt-in basis where micro frontends determine whether or not using the front end cache clients suits their needs.
```
iloveagent57, line 80: Nit: link to the frontend-app-learner-portal-enterprise repo.

docs/decisions/0004-axios-caching-implementation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+Consequences
+------------
+
+Axios is version locked to version 0.18.1 so per request overrides work until this issue gets resolved: https://github.com/RasCarlito/axios-cache-adapter/issues/99.
```
iloveagent57, line 86: Does this mean the per-request overrides (like `{ useCache: true }`) won't work on later versions?

iloveagent57 (commented):
Looking good, I have a handful of suggestions for clearer wording, and a couple of questions that I think this ADR should address.

docs/decisions/0004-axios-caching-implementation.rst
kellmarie: They could be used in conjunction if each MFE decides they want to implement them as well. The main idea was that having this caching strategy be in frontend-platform makes it widely available and accessible to MFEs instead of deciding that each MFE needs to implement their own solution/reimplement another solution copied from somewhere else

docs/decisions/0004-axios-caching-implementation.rst
kellmarie: That's the service function for getting the httpClient. The per-request overrides are the GET, POST, ETC (like `client.get('https://google.com', {cache: {maxAge: 1000}})` requests made using the client instance. Overrides like deciding to use a different storage or the maxAge of the cache data. Here's all of the options: https://www.npmjs.com/package/axios-cache-adapter#options

I think the issue is fixed or resolved in Axios 0.20 so I think at worst we might just need to skip versions between 0.18.1 and 0.20

docs/decisions/0004-axios-caching-implementation.rst
kellmarie: For the first bulletpoint the concern with authentication specific requests is data that's tied to the authenticated user still being used after the user logs out. Which if it happens that a user is still seeing their data after logging out it would only be for about 5 minutes.

I'll try figuring out a better way to word that

docs/decisions/0004-axios-caching-implementation.rst
iloveagent57: Cool, could you add some version of those two sentences to this section? :)

docs/decisions/0004-axios-caching-implementation.rst
iloveagent57: 👍 We should probably make a note about that in here or in `package.json`.

PR #134, opened 2020-12-21
feat: Adds initializeMockApp for component testing

davidjoy:
The initializeMockApp function will setup mock versions of the analytics, auth, and logging services, see function description for more details.  The i18n service is set up as normal, since it makes no requests and is suitable for tests.

Using this should save a a bunch of manual test setup to mock logging/analytics calls and set up mock HTTP clients.

```
import { initializeMockApp } from '@edx/frontend-platform/testing';
import { logInfo } from '@edx/frontend-platform/logging';

describe('initializeMockApp', () => {
  it('mocks things correctly', () => {
    const { loggingService } = initializeMockApp();
    logInfo('wut', {});
    expect(loggingService.logInfo).toHaveBeenCalledWith('wut', {});
  });
});

```

### DIFF
--- a/src/analytics/interface.js
+++ b/src/analytics/interface.js
@@ -29,9 +29,7 @@ import PropTypes from 'prop-types';
 
 const optionsShape = {
   config: PropTypes.object.isRequired,
-  httpClient: PropTypes.shape({
-    post: PropTypes.func.isRequired,
-  }).isRequired,
+  httpClient: PropTypes.func.isRequired,
   loggingService: PropTypes.shape({
     logError: PropTypes.func.isRequired,
     logInfo: PropTypes.func.isRequired,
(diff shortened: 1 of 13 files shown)


raq929 (approved):
I love it and I needed it last week. 🎉

adamstankiewicz (approved):

davidjoy:
Hmm, maybe we can pick this up again. I think I have some more code locally that I need to add...

src/analytics/interface.js
```diff
@@ -29,9 +29,7 @@ import PropTypes from 'prop-types';
 
 const optionsShape = {
   config: PropTypes.object.isRequired,
-  httpClient: PropTypes.shape({
-    post: PropTypes.func.isRequired,
-  }).isRequired,
+  httpClient: PropTypes.func.isRequired,
```
davidjoy, line 32: Somewhat confuses me, but turns out this is a `func` not an object.  Not sure why I didn't see a warning earlier...

raq929 (approved):
LGTM!

src/analytics/interface.js
davidjoy: We had a test that mocked it as an object, which I think is where this came from.

adamstankiewicz (approved):

PR #140, opened 2021-01-25
fix: resolve ErrorPage styles

adamstankiewicz:
Since we began using react-bootstrap passthru components in Paragon ~v10, the `ErrorPage` button text has been broken, containing no button text due to `Button` component props API changing.

This PR resolves that.

### Open edX

image

### edX.org

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,9 @@
   "devDependencies": {
     "@commitlint/cli": "8.2.0",
     "@commitlint/config-angular": "8.2.0",
-    "@edx/frontend-build": "5.3.2",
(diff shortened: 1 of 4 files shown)


abutterworth (approved):

package.json
```diff
@@ -70,7 +71,7 @@
     "universal-cookie": "4.0.4"
   },
   "peerDependencies": {
-    "@edx/paragon": "^7.1.3",
+    "@edx/paragon": "^12.8.0",
```
davidjoy, line 74: Is it true that this is now the minimum paragon version that frontend-platform is compatible with?

PR #144, opened 2021-01-28
Docs updates

raq929:
Updates docs to better reflect what's in the code. 😄

### DIFF
--- a/src/auth/interface.js
+++ b/src/auth/interface.js
@@ -171,7 +171,7 @@ export function redirectToLogin(redirectUrl) {
  *
  * ```
  * const url = getLogoutRedirectUrl('http://localhost/mypage');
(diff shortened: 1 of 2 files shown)


src/auth/interface.js
```diff
@@ -171,7 +171,7 @@ export function redirectToLogin(redirectUrl) {
  *
  * ```
  * const url = getLogoutRedirectUrl('http://localhost/mypage');
- * console.log(url); // http://localhost/logout?next=http%3A%2F%2Flocalhost%2Fmypage
+ * console.log(url); // http://localhost/logout?redirect_url=http%3A%2F%2Flocalhost%2Fmypage
```
adamstankiewicz, line 174: For context, both the `next` and `redirect_url` query params handle redirection after logout.

adamstankiewicz (approved):

davidjoy:
Late to the party, but thanks @raq929. 😁

PR #145, opened 2021-02-03
fix: setting paragon peer dependency back to 7.1.3

davidjoy:


davidjoy:
This is the wrong fix - closing.  Paragon must be at least 10.x

PR #146, opened 2021-02-04
fix: set the paragon peer dep to 12.8.0 to <14.0.0

davidjoy:
We’re seeing issues in MFEs running paragon 13 where they can’t build in GoCD.

See version range rules here: https://docs.npmjs.com/cli/v6/using-npm/semver

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "universal-cookie": "4.0.4"
   },
   "peerDependencies": {
-    "@edx/paragon": "^12.8.0",
+    "@edx/paragon": ">=12.8.0 <14.0.0",
     "prop-types": "^15.7.2",
     "react": "^16.9.0",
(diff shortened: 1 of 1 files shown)


abutterworth (approved):

package.json
```diff
@@ -71,7 +71,7 @@
     "universal-cookie": "4.0.4"
   },
   "peerDependencies": {
-    "@edx/paragon": "^12.8.0",
+    "@edx/paragon": ">=12.8.0 <14.0.0",
```
adamstankiewicz, line 74: I believe we could get away with `>=10.0.0 <14.0.0` as the Paragon upgrade in this library only needs when we first introduced the react-bootstrap passthru for the `Button` component, which was introduced on Paragon@10.0.0.

PR #148, opened 2021-02-06
feat: inject analytics service from frontend-platform into paragon via ``React.Context``

adamstankiewicz:
TODO


src/initialize.js
```diff
@@ -259,7 +264,12 @@ export async function initialize({
 
     // Application Ready
     await handlers.ready();
-    publish(APP_READY);
+
+    // Configure dependency injections
+    const dependencyInjections = {
+      analytics: getAnalyticsService(),
```
adamstankiewicz, line 270: TODO: only pass `sendTrackEvent` here?

src/react/PageRoute.jsx
```diff
@@ -22,7 +22,7 @@ export default function PageRoute(props) {
     if (match) {
       sendPageEvent();
     }
-  }, [match]);
+  }, [JSON.stringify(match)]);
```
adamstankiewicz, line 25: We are currently logging duplicate page events, so adding `JSON.stringify` here will make sure this effect only runs when something within `match` changes, not when the object reference changes.

example/index.jsx
```diff
@@ -10,21 +10,34 @@ import {
...
-      <PageRoute exact path="/" component={ExamplePage} />
-      <PageRoute
-        exact
-        path="/error_example"
-        component={() => <ErrorPage message="Test error message" />}
-      />
-      <AuthenticatedPageRoute exact path="/authenticated" component={AuthenticatedPage} />
+      <ParagonProvider
+        analytics={{
+          sendTrackEvent: (eventName, eventPoperties) => analyticsService.sendTrackEvent(eventName, eventPoperties),
```
davidjoy, line 29: Getting this from analyticsService avoids the interface... I think if you:

```
import { sendTrackEvent } from '@edx/frontend-platform/analytics';
...
<ParagonProvider
  analytics={{
    sendTrackEvent
  }}
>
```

That would simplify things and have it go through the interface, yah?

PR #150, opened 2021-02-17
draft: feat: Log errors as Info if they are in QUIET_ERROR_REGEXES

kdmccormick:
Just looking for initial feedback on this idea right now. This still needs tests and comments.

The idea is that individual MFEs could configure a list of `QUIET_ERROR_REGEXES`, delimited by semicolons (I considered comma-delimited, but that seems more likely to clash with regexes that folks might want to write. Maybe double-semicolons would be an even stronger delimiter?). If an error message matches any of these regexes, it will be passed to `logInfo` instead of `logError`.

Related to [TNL-7924](https://openedx.atlassian.net/browse/TNL-7924)

### DIFF
--- a/src/react/ErrorBoundary.jsx
+++ b/src/react/ErrorBoundary.jsx
@@ -1,9 +1,19 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
-import { logError } from '../logging';
+import { logError, logInfo } from '../logging';
+import { getConfig } from '../config';
 
 import ErrorPage from './ErrorPage';
+
+function shouldQuietError(error) {
+  if (!('QUIET_ERROR_REGEXES' in getConfig())) {
+    return false;
+  }
+  const quietErrorRegexes = getConfig().QUIET_ERROR_REGEXES.split(';');
+  const errorString = error.toString();
+  return quietErrorRegexes.some(regex => regex.matches(errorString));
+}
 
 /**
  * Error boundary component used to log caught errors and display the error page.
@@ -23,7 +33,8 @@ export default class ErrorBoundary extends Component {
   }
 
   componentDidCatch(error, info) {
-    logError(error, { stack: info.componentStack });
+    const log = shouldQuietError(error) ? logInfo : logError;
+    log(error, { stack: info.componentStack });
   }
 
   render() {


davidjoy (commented):
Hm, I'm thinking about whether we might want to transform errors to info at the logging service level, rather than in this ErrorBoundary component.  

I.e., in src/logging/interface.js we could check QUIET_ERROR_REGEXES inside `logError` and delegate to `logInfo` instead if it's an error we should quiet.  It could also maybe be done in NewRelicLoggingService itself, but we should think about whether it's a piece of functionality we want all logging services to have, vs. something New Relic-specific.

That would let it apply to all attempts at logging errors, not just ones that happen to bubble to this ErrorBoundary.  This ErrorBoundary only gets used if they're not caught somewhere else, and there are presumably other places in our code where we're calling `logError` directly. 

For instance, I _think_ the auth errors we see a lot of don't make it to this boundary, but get logged inside the auth service itself.

My other thought is about whether QUIET_ERROR_REGEXES should just be an array of regexes so we don't have any string parsing/manipulation issues.

Making it an array would mean it could be configured in the app's call to `initialize` by merging in extra config, but it would mean it couldn't be set in our `.env` files.  I'm not quite sure which is better.  Setting it in .env means that Open edX instances could use a different set of regexes, and that we'd have to configure it in edx-internal ourselves.  I wish it didn't have to be a parsed string though; our current environment variable configuration mechanism sucks!

robrap:
Thoughts I've had about this:
* I agree with @davidjoy that adding it to the logging service will make it more universal.
* Once the dust settles, please ensure there are docs that go along with this with example usage.
* I had also been thinking about a concept I referred to as "known errors", where the error would still be logged as an error but would get a "knownError" custom attribute. This would rely on special NRQL to ignore "known errors", but the NRQL could be written once and would apply to any newly configured errors. **NOTE:** It may be that we offer both alternatives at some point, and you can continue to implement the quiet alternative if it is your initial preference. Mostly, the "quiet" solution loses (most of) the stack details and any other diagnostic help New Relic provides for errors.
* For the quiet solution, we could add a "quietError" custom attribute that would help find the log messages.
* For either solution, I'd like the docs to recommend considering an alert (or alerts) that ensure that that the quiet or known errors don't get out of hand in a way you didn't expect. Could be one alert for the custom attribute.
* I had considered using New Relic's "LIKE" syntax over regex so that someone could quickly go from a NRQL query that works and copy that in. However, since logging should work against things other than New Relic, a better solution might be adding into the doc how to quickly convert from LIKE syntax to regex, which would be pretty simple. LIKE is case-insensitive, but that is probably not a big deal.
* I was wondering if we'd eventually want to compare against other things like the route? This might apply to edx-platform and the backend more than MFEs, because MFEs might already be a small enough scope. I agree the config story for the MFEs isn't great, but I like that you aren't blocking yourself on fixing that.
* Related to the last point, the error string should be pretty stable, but I did want to improve some of the axios error strings at some point which might affect regex for those.  We could add a note to the docs for these.

robrap:
UPDATE: See https://github.com/edx/edx-platform/pull/26980/files for backend implementation.

robrap:
FYI: @iloveagent57 @binodpant: This is the PR I had mentioned.

kdmccormick:
Closing in favor of https://github.com/edx/frontend-platform/pull/169

PR #154, opened 2021-03-05
fix: AA-513: Reset segment state if anon user and there is a segment user id

Dillon-Dumesnil:
If we are seeing an anonymous user, but the segment user id is still
set, we believe the segment user id is coming from a different user on
the same machine. This will make sure we clear out that storage and
then the indentify call will make a new anonymous id

### DIFF
--- a/src/analytics/SegmentAnalyticsService.js
+++ b/src/analytics/SegmentAnalyticsService.js
@@ -153,10 +153,34 @@ class SegmentAnalyticsService {
    * Send anonymous identify call to Segment's identify.
    *
    * @param {*} [traits]
+   * @returns {Promise} Promise that will resolve once the document readyState is complete
    */
   identifyAnonymousUser(traits) {
-    global.analytics.identify(traits);
-    this.hasIdentifyBeenCalled = true;
+    // if we do not have an authenticated user (indicated by being in this method),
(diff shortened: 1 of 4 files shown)


src/analytics/SegmentAnalyticsService.js
```diff
@@ -155,8 +155,17 @@ class SegmentAnalyticsService {
    * @param {*} [traits]
    */
   identifyAnonymousUser(traits) {
-    global.analytics.identify(traits);
-    this.hasIdentifyBeenCalled = true;
+    // if we do not have an authenticated user (indicated by being in this method),
+    // but we still have a user id associated in segment, reset the local segment state
+    // This has to be wrapped in the analytics.ready call because the analytics.user()
+    // function isn't available until analytics is loaded.
+    global.analytics.ready(() => {
```
davidjoy, line 162: So I think perhaps what we should do is make `identifyAnonymousUser` return a promise so we can `await` it in the `initialize` function.  That way we don't proceed with initialization without having finished identifying the user - assuming that `global.analytics.ready` is asynchronous.  Then in initialize, we can just call `await identifyAnonymousUser` and all will be good.

davidjoy (approved):

davidjoy (approved):

PR #156, opened 2021-03-15
Early failure message when backend is down.

adeelkhan:
WIP:
This patch would enable MFE to fail early just in
case of backend (LMS) is down and it want to show some user
facing page instead of showing a blank page.

VAN-129

### DIFF
--- a/src/auth/AxiosJwtAuthService.js
+++ b/src/auth/AxiosJwtAuthService.js
@@ -18,6 +18,7 @@ const optionsPropTypes = {
     REFRESH_ACCESS_TOKEN_ENDPOINT: PropTypes.string.isRequired,
     ACCESS_TOKEN_COOKIE_NAME: PropTypes.string.isRequired,
     CSRF_TOKEN_API_PATH: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


davidjoy:
Some initial thoughts - 

1. Can you describe in a bit more detail the situation this is addressing?  I'm not sure I understand what issues we're running into.
2. I think if we provide an option for this, it should likely be an option on the `initialize` function which can be passed in when the application is started up.  If an application is providing some alternate handling, then that won't be changing on a per-deployment basis.  Environment variables are useful for when we want to change behavior depending on the environment (dev, prod, Open edX instances, etc.), but I don't think that's the case here.
3. SHOULD_RETRY is pretty vague; if we do add an option, it should be unambiguous.

abdullahwaheed:
Hi @adeelkhan! This PR is inactive for quite a time now, could you please resolve the conversation and update your PR

openedx-webhooks:
Although this pull request is already closed, I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #162, opened 2021-04-14
feat: add email from JWT into authenticatedUser object

adeelehsan:


waheedahmed (commented):
Add feat at the start of the commit message otherwise, semantic-release will not pick these changes. Also, rephrase the commit message to something like `feat: add email in authenticated user`.

davidjoy (approved):

PR #164, opened 2021-04-21
fix: prevent JS error when calling analytics.user()

adamstankiewicz:


src/analytics/interface.test.js
```diff
@@ -89,6 +89,14 @@ describe('analytics identifyAuthenticatedUser', () => {
 });
 
 describe('analytics identifyAnonymousUser', () => {
+  beforeEach(() => {
```
Dillon-Dumesnil, line 92: Probably don't need this to be in a `beforeEach`, right? Just needs to be defined once. Or maybe put up next to the other mocks (lines 45-48)

Dillon-Dumesnil (approved):
Thanks for catching and fixing!

.env.development
```diff
@@ -14,7 +14,7 @@ STUDIO_BASE_URL=http://localhost:18010
 MARKETING_SITE_BASE_URL=http://localhost:18000
 ORDER_HISTORY_URL=http://localhost:1996/orders
 REFRESH_ACCESS_TOKEN_ENDPOINT=http://localhost:18000/login_refresh
-SEGMENT_KEY=ul
+SEGMENT_KEY=null
```
davidjoy, line 17: 😩

src/analytics/SegmentAnalyticsService.js
```diff
@@ -158,28 +158,17 @@ class SegmentAnalyticsService {
   identifyAnonymousUser(traits) {
     // if we do not have an authenticated user (indicated by being in this method),
     // but we still have a user id associated in segment, reset the local segment state
-    // This has to be wrapped in the document.readyState logic because the analytics.user()
+    // This has to be wrapped in the analytics.ready() callback because the analytics.user()
     // function isn't available until the analytics.js package has finished initializing.
     return new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
-      if (document.readyState === 'complete') {
+      global.analytics.ready(() => {
```
davidjoy, line 164: The code here was written to account for situations where the analytics.ready() event had already been fired by the time identifyAnonymousUser was called.  Do we have a guarantee that global.analytics.ready() is called immediately if analytics is already in a ready state?

davidjoy (commented):
Sanity check.

src/analytics/SegmentAnalyticsService.js
davidjoy: (Otherwise, this will effectively do nothing if analytics is already "ready", right?)

adamstankiewicz: @davidjoy Good question. It looks like `ready` is implemented such that if it is already in a ready state, the callback function is immediately called: https://github.com/segmentio/analytics.js/blob/master/analytics.js#L678-L687

To further confirm this, I did something like the following:

```jsx
global.analytics.ready(() => {
  console.log('analytics.js is ready');
});

setTimeout(() => {
  global.analytics.ready(() => {
    console.log('ready callback still gets called');

    if (global.analytics.user().id()) {
      global.analytics.reset();
    }
    global.analytics.identify(traits);
    this.hasIdentifyBeenCalled = true;
    resolve();
  });
}, 5000);
```

where both `console.log` lines are called.

davidjoy: Perfect!

PR #167, opened 2021-04-26
docs: include info about the example app and a PR template

adamstankiewicz:
Updates the README to explicitly mention the example app included in this repo so that others may be more aware of it existing to QA against.


iloveagent57 (approved):
🏆 Nice!

davidjoy (approved):

PR #169, opened 2021-05-06
feat: Support classifying certain errors as ignored errors in an MFE.

doctoryes:
Ignored errors are sent to New Relic as page actions instead of JS errors,
allowing those errors to still be tracked as occurring but without causing
unnecessary alerts.

Ignored errors are configured per-MFE, *not* globally.

https://openedx.atlassian.net/browse/TNL-7924

Related PR: 
- https://github.com/edx/edx-internal/pull/5046
- https://github.com/edx/frontend-app-learning/pull/447

**Merge checklist:**

- [x] Consider testing your code modifications in another local micro-frontend using local aliases configured via [the `module.config.js` file in `frontend-build`](https://github.com/edx/frontend-build#local-module-configuration-for-webpack).

**Post merge:**


src/logging/NewRelicLoggingService.js
```diff
@@ -77,14 +112,26 @@ export default class NewRelicLoggingService {
...
+        Separate the errors into expected errors and unexpected errors.
+        Expected errors are logged via adding a page action.
+        Unexpected errors are logged via noticeError and count as "JS Errors" for the application.
+    */
+    /* for (const expectedErr of this.expectedErrors) { */
+    let expected = false;
+    this.expectedErrors.forEach(expectedErr => {
+      if (!expected
+          && ((typeof error === 'string' && error.includes(expectedErr))
+          || (error.message && error.message.includes(expectedErr)))) {
```
doctoryes, line 126: Currently, this code only supports substring detection. Should it instead support regex expressions?

kdmccormick: I think regex would be good, especially since one the expected errors messages is literally `"Script error"` -- wouldn't want to match every message with that substring!

kdmccormick: JS tip -- [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is a nice way to do the check-against-event-array-element loop idiom here.

src/logging/NewRelicLoggingService.js
```diff
@@ -17,18 +17,47 @@ function fixErrorLength(error) {
...
+    window.newrelic.addPageAction(actionName, { message, ...customAttributes });
+  }
+}
+
+function logUnexpectedError(error, customAttributes) {
+  if (process.env.NODE_ENV === 'development') {
+    console.error(error, customAttributes); // eslint-disable-line
+  }
+  if (window && typeof window.newrelic !== 'undefined') {
+    // Note: customProperties are not sent.  Presumably High-Security Mode is being used.
```
kdmccormick, line 38: could you explain this more? are customProperties different than customAttributes?

kdmccormick (commented):
General approach LGTM.

I does differ in some ways from how edx-platform expected error are being handled: https://github.com/edx/edx-platform/pull/26980. But, NR APM is different than NR Browser, and Python is different from JS. So the differences seem reasonable to me.

I would defer to @robrap for a more in-depth opinion, though. He'd expressed wanting to be looped in when we got around to this ticket.

doctoryes (commented):
Thanks for the initial review!

robrap:
@doctoryes: If you are ok with matching the language (and potentially the functionality) of the backend work, we'd have the following:
* ignored errors: This is like what you have implemented, where we don't actually keep the error/stacktrace information in New Relic.
* expected errors: This is where we kept the error/stacktrace in New Relic, but we added an `error_expected` custom attribute. If implemented, this would leave the `logError` call, but send the new custom attribute.

Choosing to ignore vs expect an error would mostly be a function of whether you want the full stacktrace info, etc. In both cases, custom NRQL could be written to include or exclude the correct data using the custom attributes.

I think it would be unnecessarily confusing to call this "expected errors", but have a different meaning for "expected errors" on the backend. At a minimum, would you be ok with calling these "ignored errors"?

Let me know your thoughts.  Happy to discuss synchronously too if that would be simpler.

doctoryes:
Ah - I'd mis-read your PR, @robrap . I'm fine with calling what's been implemented so far ignored errors. I'm not sure yet whether we'll need to implement expected errors - but if I do, I'll mirror the back-end pattern. Your message above was a helpful summary - thanks!

doctoryes:
@robrap : You may know the answer to this New Relic question: Does NR for edx.org have high-security mode (aka enterprise security mode) enabled? If so, it appears that custom attributes wouldn't be passed in `noticeError()` calls, according to this page:
https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode/

robrap:
@doctoryes: I'm pretty sure we are **not** using high-security mode, and that we can pass custom attributes. You can see some examples here, https://one.nr/0xZw0o3q3Qv, like all the `httpErrorXxx` attributes on some of the results.

**Note:** It looks like `message` contains the custom attributes data, and in some cases, is the only place the custom attributes are captured. This doesn't seem right. It's unclear to me if our `frontend-logging` code is at fault for this, or incorrect usage of the `logInfo` call in places, or both. Either way, looks like some bugs.

robrap:
@doctoryes: Reminder that the old "expected error" terminology is still used in the PR title and description.

doctoryes:
I tested this change in the frontend-app-learning MFE and verified that a configured regular expression would match an intended error and be sent as a page action instead of a JS error.

src/logging/NewRelicLoggingService.js
```diff
@@ -17,18 +17,46 @@ function fixErrorLength(error) {
   return error;
 }
 
+/* Constants used as New Relic page action names. */
+const pageActionNameInfo = 'INFO';
+const pageActionNameIgnoredError = 'IGNORED_ERROR';
+
+function sendOther(actionName, message, customAttributes) {
```
robrap, line 24: Nit: Maybe call this `sendPageAction`? Or, `sendNonError`? The word 'Other` doesn't contain enough info (i.e. other than what?).

src/logging/NewRelicLoggingService.js
```diff
@@ -17,18 +17,46 @@ function fixErrorLength(error) {
...
- * When you use `logError`, your errors will appear under "JS errors"
- * for your Browser application.
+ * When you use `logError`, your errors will be checked to see if they're ignored *or* not.
+ * Not-ignored errors will appear under "JS errors" for your Browser application.
  *
  * ```
  * SELECT * from JavaScriptError WHERE errorStatus is not null SINCE 10 days ago
  * ```
  *
+ * while ignored errors will appear in New Relic Insights as page actions, which can be queried:
```
robrap, line 54: Nit:
```suggestion
 * Ignored errors will appear in New Relic Insights as page actions, which can be queried:
```

src/logging/NewRelicLoggingService.js
```diff
@@ -45,6 +73,20 @@ function fixErrorLength(error) {
...
+  constructor(options) {
+    this.config = options ? options.config : undefined;
+    /*
+        Array of explicit error message regexes. If an error message matches a regex, the error is considered
+        an *ignored* error and submitted to New Relic as a page action - not an error.
+
+        Ignored error regexes are configured per frontend application.
+        Add new error message regexes in edx-internal YAML as needed, ordered roughly by occurrence frequency.
+        The list of error regexes is represented in the .env files as a string, with multiple regexes delimited
+        by three pipes '|||'.
```
robrap, line 85: 1. Call out documentation that is specific to edx.org.
```suggestion

        For edx.org, add new error message regexes in edx-internal YAML as needed, ordered roughly by
        occurrence frequency. The list of error regexes is represented in the .env files as a string, with
        multiple regexes delimited by three pipes '|||'.
```
2. Is the three pipe delimiter true for everyone? Any parts of this that aren't edx.org specific, maybe pull out and up.
3. Consider providing an example regex.

src/logging/interface.js
```diff
@@ -5,13 +5,19 @@
  * NewRelicLoggingService that sends messages to NewRelic that can be seen in NewRelic Browser and
  * NewRelic Insights. When in development mode, all messages will instead be sent to the console.
  *
- * When you use `logError`, your errors will appear under "JS errors"
- * for your Browser application.
+ * When you use `logError`, your errors will be checked to see if they're ignored *or* not.
```
robrap, line 8: Could we have docs just in NewRelicLoggingService.js to avoid duplication, and reference them from here? Also, note that these docs are partially specific to the New Relic implementation.

src/logging/NewRelicLoggingService.js
doctoryes: Agreed - I meant for this to be a TBD name. Will change.

src/logging/NewRelicLoggingService.js
doctoryes: Will do.

src/logging/NewRelicLoggingService.js
doctoryes: 1. Will do.
2. Yes, that delimiter is intended to be for everyone.
3. Will do.

src/logging/interface.js
doctoryes: Fine with me - will do.

kdmccormick (commented):
Great work! Pending Robert's comments, this LGTM.

doctoryes:
Ready for re-review - added some solid examples.

src/logging/NewRelicLoggingService.js
```diff
@@ -45,6 +73,31 @@ function fixErrorLength(error) {
...
 export default class NewRelicLoggingService {
+  constructor(options) {
+    this.config = options ? options.config : undefined;
+    /*
+        Array of explicit error message regexes. If an error message matches a regex, the error is considered
+        an *ignored* error and submitted to New Relic as a page action - not an error.
+
+        Ignored error regexes are configured per frontend application (MFE).
+
+        For edx.org, add new error message regexes in edx-internal YAML as needed, ordered roughly by
```
robrap, line 84: Non-blocking things to consider:
1. I still think the "For edx.org..." sentence should be moved to the bottom of the docstring. The second sentence here might be ignored outside of edx.org, but it is meant for everyone.
2. Is "ordered roughly by occurrence frequency" due to potential performance concerns? I don't think I'd worry about it, and if you are concerned, couldn't all the regex's be combined into a single regex?

robrap (approved):
Looks good @doctoryes. Thank you.

src/logging/NewRelicLoggingService.js
robrap: Sorry if the answer is obvious, but why did we use the special delim and not just have a single regex? Is it because the regex might get too complicated if a specific error also needed to use `|`? For example, it could have just been:
```
>>> x = re.compile(r"(^\[frontend-auth\] Axios Error|Loading chunk [\d]+ failed. \(error: https://learning.edx.org/[\d\w.]+\.js\))")
>>> x.search('hello')
>>> x.search('[frontend-auth] Axios Error (Request): Network Error (possible local connectivity issue)')
<re.Match object; span=(0, 27), match='[frontend-auth] Axios Error'>
>>> x.search('Loading chunk 8 failed. (error: https://learning.edx.org/8.00acf25bc2384f18cf76.js)')
<re.Match object; span=(0, 83), match='Loading chunk 8 failed. (error: https://learning.>
```
Note: There were some minor issues with your example regex, like some missing or unwanted `\` characters.

doctoryes: There's actually no reason why a single regex can't be used here. It would simplify the code in places, which is another plus. The regex itself won't seem any more complex than it currently is - given the `dotenv` restriction of a single string per config env var. I'm going to make this change throughout my three PRs - thanks!

The example regex does need updating - agreed. Here's the example in JS - not Python with its fancy `r''` operator 😸 :
```
> x = new RegExp('^\\[frontend-auth\\] Axios Error|Loading chunk [\\d]+ failed. \\(error: https://learning.edx.org/[\\d\\w\\.]+.js\\)');
/^\[frontend-auth\] Axios Error|Loading chunk [\d]+ failed. \(error: https:\/\/learning.edx.org\/[\d\w\.]+.js\)/
> 'Loading chunk 8 failed. (error: https://learning.edx.org/8.00acf25bc2384f18cf76.js)'.match(x);
[
  'Loading chunk 8 failed. (error: https://learning.edx.org/8.00acf25bc2384f18cf76.js)',
  index: 0,
  input: 'Loading chunk 8 failed. (error: https://learning.edx.org/8.00acf25bc2384f18cf76.js)',
  groups: undefined
]
> 'Loading chunk 4 failed. (error: https://learning.edx.org/4.ebd72a4f92c5e7a0214b.js)'.match(x);
[
  'Loading chunk 4 failed. (error: https://learning.edx.org/4.ebd72a4f92c5e7a0214b.js)',
  index: 0,
  input: 'Loading chunk 4 failed. (error: https://learning.edx.org/4.ebd72a4f92c5e7a0214b.js)',
  groups: undefined
]
> '[frontend-auth] Axios Error (Request): Network Error (possible local connectivity issue) post https://courses.edx.org/login_refresh'.match(x);
[
  '[frontend-auth] Axios Error',
  index: 0,
  input: '[frontend-auth] Axios Error (Request): Network Error (possible local connectivity issue) post https://courses.edx.org/login_refresh',
  groups: undefined
]
```
I attempted a `string.replace()` regex to avoid the user having to escape characters themselves - but I couldn't find one that worked in all cases. For example, is the `.` meant to be a literal `.` - or to match any character? I decided it's best to leave those decisions in the hands of the regex creator.

robrap: I agree that you don't want to determine where there should be a `\` and where there shouldn't be. However, your example and the actual setting should  not need a `\\` to represent a `\`, right? Or, does it need that to get through the config code?

doctoryes: It's apparently got to do with the way JS does string escaping - but I'm not entirely sure why.

My latest commit adds actual JS code which a developer can run to test out their regex additions. Can you please run that and see if the escaping makes the matches occur properly? Thanks.

robrap: Looks great @doctoryes.

Consider adding something like:
```
'Example message that should not match'.match(x)
```

Still has 👍 from me.

robrap: @doctoryes: Also, for the PR where you are setting this, consider a comment in yml with:
1. Link to instructions, 
2. Example error messages that match each section of the regex.

src/logging/NewRelicLoggingService.js
```diff
@@ -45,6 +73,39 @@ function fixErrorLength(error) {
...
+        The list of error regexes is represented in the .env files as a string, with multiple regexes enabled
+        via the '|' regex syntax.
+
+        For example, here's a .env line which ignores two specific errors:
+
+        IGNORED_ERROR_REGEXES='^\\[frontend-auth\\] Axios Error|Loading chunk [\\d]+ failed. \\(error: https://learning.edx.org/[\\d\\w\\.]+.js\\)'
+
+        This example would ignore errors with the following messages:
+
+        Loading chunk 8 failed. (error: https://learning.edx.org/8.00acf25bc2384f18cf76.js)
```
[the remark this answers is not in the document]
doctoryes: I wanted to add a more complicated regex just to give breadcrumbs for anyone who wanted to add something more complex/specific. I do *not* plan to ignore this specific error - it's just used as an example.

src/logging/NewRelicLoggingService.js
doctoryes: I added the non-matching example. And I'll add a link to the instructions here to the YML after this PR merges.

src/logging/NewRelicLoggingService.js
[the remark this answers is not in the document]
robrap: Personally, I'd err on having less complex examples (people can research regex elsewhere) in favor of not misleading someone into thinking this is an error that they should consider ignoring, when they probably shouldn't.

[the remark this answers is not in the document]
doctoryes: Done.

PR #174, opened 2021-05-14
fix: Corrects some of the docstring text from IGNORED_ERROR_REGEX.

doctoryes:
**Description:**

This PR was created mainly to force a patch version bump and a new release of frontend-platform, after the previous `chore:` commit which was really more of a feature. But it also sharpens the docstring.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/src/logging/NewRelicLoggingService.js
+++ b/src/logging/NewRelicLoggingService.js
@@ -86,7 +86,7 @@ export default class NewRelicLoggingService {
 
         For example, here's a .env line which ignores two specific errors:
 
-        IGNORED_ERROR_REGEXES='^\\[frontend-auth\\] Unimportant Error|Specific non-critical error #[\\d]+'
+        IGNORED_ERROR_REGEX='^\\[frontend-auth\\] Unimportant Error|Specific non-critical error #[\\d]+'
 
         This example would ignore errors with the following messages:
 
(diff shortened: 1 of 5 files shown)


src/logging/NewRelicLoggingService.js
```diff
@@ -76,13 +76,13 @@ export default class NewRelicLoggingService {
...
-        Array of explicit error message regexes. If an error message matches a regex, the error is considered
-        an *ignored* error and submitted to New Relic as a page action - not an error.
+        String which is an explicit error message regex. If an error message matches the regex, the error
+        is considered an *ignored* error and submitted to New Relic as a page action - not an error.
 
         Ignored error regexes are configured per frontend application (MFE).
 
-        The list of error regexes is represented in the .env files as a string, with multiple regexes enabled
-        via the '|' regex syntax.
+        All ignored error regexes are represented in the .env files as a single string, with multiple regexes
```
robrap, line 84: I find the term "regexes" confusing, because there is only a single regex.  What about something like:
```
The regex for all ignored errors are represented in the .env files as a single string. If you need to
ignore multiple errors, use the standard `|` regex syntax.
```

doctoryes: Done. Please re-review.

kdmccormick (commented):
Given the change from array-of-regexes to a single union regex, would it make more sense to call it `IGNORE_ERROR_REGEX`?

doctoryes:
@kdmccormick Makes sense - done!

kdmccormick (approved):
:rocket:

PR #204, opened 2021-08-06
feat(auth): allow forceRefresh in fetchAuthenticatedUser

waheedahmed:
**Description:**
Allow an optional parameter in `fetchAuthenticatedUser` that forces it to refresh the JWT token even if the token is not yet expired.

VAN-682

**Merge checklist:**

- [x] Consider testing your code modifications in another local micro-frontend using local aliases configured via [the `module.config.js` file in `frontend-build`](https://github.com/edx/frontend-build#local-module-configuration-for-webpack).

**Post merge:**

### DIFF
--- a/src/auth/AxiosJwtAuthService.js
+++ b/src/auth/AxiosJwtAuthService.js
@@ -208,6 +208,8 @@ class AxiosJwtAuthService {
         roles: decodedAccessToken.roles || [],
         administrator: decodedAccessToken.administrator,
       });
+    } else {
+      this.setAuthenticatedUser(null);
     }
 
     return this.getAuthenticatedUser();
--- a/src/auth/AxiosJwtAuthService.js
+++ b/src/auth/AxiosJwtAuthService.js
@@ -197,8 +197,8 @@ class AxiosJwtAuthService {
    * @returns {Promise<UserData>|Promise<null>} Resolves to the user's access token if they are
    * logged in.
    */
-  async fetchAuthenticatedUser() {
-    const decodedAccessToken = await this.jwtTokenService.getJwtToken();
+  async fetchAuthenticatedUser(options = {}) {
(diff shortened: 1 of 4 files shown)


src/auth/AxiosJwtAuthService.js
```diff
@@ -197,8 +197,8 @@ class AxiosJwtAuthService {
    * @returns {Promise<UserData>|Promise<null>} Resolves to the user's access token if they are
    * logged in.
    */
-  async fetchAuthenticatedUser() {
-    const decodedAccessToken = await this.jwtTokenService.getJwtToken();
+  async fetchAuthenticatedUser(options = {}) {
+    const decodedAccessToken = await this.jwtTokenService.getJwtToken(options.forceRefresh || false);
 
     if (decodedAccessToken !== null) {
```
waheedahmed, line 203: @davidjoy Do we need to unset the user here if `forceRefresh` is `true` and `decodedAccessToken` is `null`? Otherwise, it causes an extra redirect since the returned value will be a valid token.

davidjoy: Ahh, good point, yes, I think you'll want to do that.  Otherwise the app will continue to have a value returned by `getAuthenticatedUser`.  If the decoded token is `null`, I think we want to `this.setAuthenticatedUser(null)`

davidjoy:
Think we'll be able to approve and get this in fast once we fix un-setting the authenticated user if the token is null. 👍🏻

src/auth/AxiosJwtAuthService.js
waheedahmed: Done.

waheedahmed:
@davidjoy ready for review.

davidjoy (approved):
Looks good!

PR #211, opened 2021-08-17
fix(analytics): avoid errors when we bail on segment identification

mikix:
If we timeout during segment identification, stop any further segment calls from happening, as they will fail anyway. And that just generates JS error churn.

This was noticed because about 18% of the learning MFE JS errors are now "Identify must be called before other tracking events." after we recently updated to the latest `frontend-platform`. I'm not sure this code path I fixed is 100% responsible for that uptick. But it certainly *can* result in that error.

### DIFF
--- a/src/analytics/SegmentAnalyticsService.js
+++ b/src/analytics/SegmentAnalyticsService.js
@@ -187,9 +187,13 @@ class SegmentAnalyticsService {
         resolve();
       });
 
-      // if the segment domain is blocked on user's network/browser, this promis does not get
+      // if the segment domain is blocked on user's network/browser, this promise does not get
       // resolved and user see a blank page, set timeout out to resolve the promise.
-      setTimeout(resolve, 2000);
+      setTimeout(() => {
+        // Other segment calls won't work if we are not identified. So let's just un-initialize ourselves.
+        this.segmentInitialized = false;
+        resolve();
+      }, 2000);
     });
   }
 


src/analytics/SegmentAnalyticsService.js
```diff
@@ -187,9 +187,13 @@ class SegmentAnalyticsService {
...
 
-      // if the segment domain is blocked on user's network/browser, this promis does not get
+      // if the segment domain is blocked on user's network/browser, this promise does not get
       // resolved and user see a blank page, set timeout out to resolve the promise.
-      setTimeout(resolve, 2000);
+      setTimeout(() => {
+        // Other segment calls won't work if we are not identified. So let's just un-initialize ourselves.
+        this.segmentInitialized = false;
+        resolve();
+      }, 2000);
```
mikix, line 196: How bad does it screw with our metrics if we don't have segment events for people that timeout at two seconds? How many people do we think hit this? We saw 11.6k occurances of the segment error in 5 days. Which I guess isn't enormous. Does that number feel right?

davidjoy: I'm not a fan of using arbitrary timeouts like this, honestly... I didn't get a chance to chime in on the previous PR that added this timeout.  On poor connections, 2 seconds may be reasonable; seems like we're likely to drastically skew our metrics if we start excluding users with the slowest connections.

davidjoy: We're hitting this, presumably, if `global.analytics.ready` doesn't fire.  I'd like to understand the conditions under which it doesn't fire and see if we can't detect those, rather than just delaying.

davidjoy (approved):
I left a few comments.  I think this PR is an improvement as is, so I'll "Approve", but I really wish there were a better way to do this than with a timeout, and I think we're likely to be skewing our metrics now.

davidjoy:
Can we write a test for this? 😬

mikix:
> I'd like to understand the conditions under which it doesn't fire and see if we can't detect those, rather than just delaying.

Me too - but I'm not involved in all that. I'm just trying to stop 20% of learning MFE errors in prod.

> Can we write a test for this?

Love the idea. But this file has no existing tests at all, where I can hook a new one in. And as above, I'm just trying to drive-by fix a mistake.

PR #222, opened 2021-09-15
fix(analytics): events not firing for anonymous users

waheedahmed:
Also refactored the logic of resolving the promise for `identifyAnonymousUser`.

Removed the setTimeout logic and used a custom event to resolve the promise if the script has failed to load.

**Description:**

Describe what this pull request changes, and why. Include implications for people using this change.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/src/analytics/SegmentAnalyticsService.js
+++ b/src/analytics/SegmentAnalyticsService.js
@@ -87,6 +87,11 @@ class SegmentAnalyticsService {
       // Create an async script element based on your key.
       const script = document.createElement('script');
       script.type = 'text/javascript';
+      script.onerror = () => {
+        this.segmentInitialized = false;
+        const event = new Event('segmentFailed');
+        document.dispatchEvent(event);
+      };
       script.async = true;
       script.src = `https://cdn.segment.com/analytics.js/v1/${key}/analytics.min.js`;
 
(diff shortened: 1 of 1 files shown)


src/analytics/SegmentAnalyticsService.js
```diff
@@ -187,13 +192,7 @@ class SegmentAnalyticsService {
...
       });
 
-      // if the segment domain is blocked on user's network/browser, this promise does not get
-      // resolved and user see a blank page, set timeout out to resolve the promise.
-      setTimeout(() => {
-        // Other segment calls won't work if we are not identified. So let's just un-initialize ourselves.
-        this.segmentInitialized = false;
-        resolve();
-      }, 2000);
+      document.addEventListener('segmentFailed', resolve);
```
davidjoy, line 195: I would expect the segmentFailed event to fire before this listener is added.  Is it the case that you only expect this to catch the segmentFailed event if the script loading has timed out, and thus is likely to occur after identifyAnonymousUser is called?

If so, I'd add some comments to that effect, explaining that it'll only be used in certain circumstances.  Otherwise someone might delete the code thinking it doesn't function as written.

davidjoy (changes requested):
Think we need to document what's happening here, assuming I'm describing it right.  If not, there may be a race condition here?

src/analytics/SegmentAnalyticsService.js
waheedahmed: @davidjoy that was my concern as well with this but when I blocked the segment CDN domain in my browser during my testing it always worked as expected resolved the promise. And this is generally for all error cases but specifically for the blocking scenario instead of timeout.

Sure, I'll add the explaining comment.

waheedahmed: @davidjoy added the comment, kindly take another look.

davidjoy (approved):

PR #245, opened 2021-11-10
BOM-2988: Replace Travis with GitHub CI

UsamaSadiq:
**Issue:** [BOM-2988](https://openedx.atlassian.net/browse/BOM-2988)

### Description
- Replace `Travis` with `GitHub CI`.


nedbat:
This changes the release trigger from commits to tags.  @davidjoy is that in line with fedx practice?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: npm ci
+
+      - name: Create Build
+        run: npm run build
+
+      - name: Release Package
+        env:
+          GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.SEMANTIC_RELEASE_NPM_TOKEN }}
+        run: npm semantic-release
```
davidjoy, line 32: We need to pin this to `semantic-release@17` for now because v18 is not compatible with node 12.

davidjoy: Oh, also, this says `npm semantic-release` - that should be `npx semantic-release`.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Deploy to GH pages
+        env:
+          GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.SEMANTIC_RELEASE_NPM_TOKEN }}
+        run: npm deploy-storybook-ci
```
davidjoy, line 38: Not sure I understand what this line is doing - frontend-platform does not use storybook.

davidjoy:
@nedbat I'm not sure I follow the commits/tags comment... My understanding is that `semantic-release` is responsible for creating a tag for the release based on the conventional commit messages in the commits, which is what we do in all our frontend libraries.  Does this change that process?

davidjoy (changes requested):
Think I'm seeing some issues here w/r/t `npm` -> `npx` and a storybook action.  

I''l note that in frontend-build we already have a Github release workflow.  I'd suggest we keep this one consistent with that: https://github.com/edx/frontend-build/blob/master/.github/workflows/release.yml

nedbat:
@davidjoy the `.travis.yml` file had this:
```
after_success:
- npx semantic-release@17
- codecov
```
so semantic-release was run after every test run.  The new action has:
```
name: Release CI
on:
  push:
    tags:
      - '*'
```
so the Release CI action is only run when a new tag is pushed, and it's the Release CI that runs semantic-release.

davidjoy:
Oh, thank you, I didn't catch that.  That sounds like a sane change; I don't know why we'd have semantic-release run anytime except when merging into master.

nedbat:
> I don't know why we'd have semantic-release run anytime except when merging into master.

OK, except this is less frequently than that: only when someone explicitly makes and pushes a tag.

davidjoy:
Okay, now I'm confused again... semantic-release is responsible for creating the tags and does that as part of its work, after it figures out the version bump based on the commits being merged in to master.  We definitely don't want that to become a manual process, and I'm pretty sure semantic-release will fail if the tag already exists.

.github/workflows/npm-publish.yml
UsamaSadiq: Updated the workflow to use the `cycjimmy/semantic-release-action@v2` instead for release.

.github/workflows/npm-publish.yml
UsamaSadiq: Updated the workflow so these changes are not needed now.

UsamaSadiq:
@davidjoy I've updated the `Release CI` to only trigger incase of `master` branch as it was being done in `Travis` before. Is this the desired case we want?

.travis.yml
```diff
@@ -1,24 +0,0 @@
...
-node_js: 12
-script:
-- npm run lint
-- npm run test
-- npm run build
-- npm run docs
-after_success:
-- npx semantic-release@17
-- codecov
-deploy:
```
davidjoy, line 11: The frontend-platform build is a bit different than others in that it publishes the documentation site to Github Pages.  I'm not seeing this behavior preserved in the Github Actions version.

Here's an example of this being done via Github Actions in another repository: https://github.com/openedx/openedx-tech-radar/blob/main/.github/workflows/release.yml#L24

davidjoy (changes requested):
Think we're missing the Github Pages deploy.  Otherwise the release trigger sounds right.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+          semantic_version: 16
+        env:
+          GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.SEMANTIC_RELEASE_NPM_TOKEN }}
+
+      - name: Deploy to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.5
+        with:
+          branch: gh-pages # The branch the action should deploy to.
+          folder: dist # The folder the action should deploy.
```
UsamaSadiq, line 40: @davidjoy I'm not sure about the `folder` argument. Is the default `dict` value okay?

.travis.yml
UsamaSadiq: Added the GitHub Pages deployment step.

.github/workflows/release.yml
davidjoy: Yes, I think that should be correct (`dist`, not `dict`, of course, but the code is right!)

davidjoy (approved):
Lets get this in there and see if it works!

PR #296, opened 2022-03-03
fix: Use strategy matrix in CI workflow

Jawayria:
JIRA: https://openedx.atlassian.net/browse/BOM-3336

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,19 +9,19 @@ on:
       - "**"
 
 jobs:
-  build:
+  tests:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        node: [16]
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,18 +11,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        node: [16]
```
iamsobanjaved, line 16: Are we all good by just mentioning the node version here without setting up that version? Does that happen by itself when we specify this?

Jawayria: Oops! it was my mistake. Updated now

aht007 (approved):

iamsobanjaved (approved):

PR #316, opened 2022-04-13
fix: add COMMERCE_COORDINATOR_BASE_URL to config

dianekaplan:
**Description:**
Adding a base url for our new commerce-coordinator! This will enable various MFEs to import it from one place. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,6 +1,7 @@
 EXAMPLE_VAR=Example Value
 ACCESS_TOKEN_COOKIE_NAME=edx-jwt-cookie-header-payload
 BASE_URL=http://localhost:8080
+COMMERCE_COORDINATOR_BASE_URL=http://localhost:8140
 CREDENTIALS_BASE_URL=http://localhost:18150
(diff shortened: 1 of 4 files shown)


davidjoy:
In general we shouldn't add things here that aren't yet part of the Open edX platform (which doesn't mean we haven't violated that at times)

Is it maybe premature to add this URL?  What MFEs need it?

dianekaplan:
> In general we shouldn't add things here that aren't yet part of the Open edX platform (which doesn't mean we haven't violated that at times)
> 
> Is it maybe premature to add this URL? What MFEs need it?

Good to know! Yes this is very early, the very first item we'll be testing out for ourselves. The first will be frontend-app-ecommerce, later frontend-app-payment, etc. Do you recommend we just add locally/individually into those repos instead of pulling from here with everything else?

davidjoy:
I think for now I'd just add it in those yeah. And we can add it here later when it becomes more official/supported.

PR #334, opened 2022-05-17
feat: support conversation to/from snakecase for http clients

long74100:
**Description:**

Resolves https://github.com/openedx/frontend-wg/issues/94.

Adds a way to add middleware when calling `getAuthenticatedHttpClient` or `getHttpClient`, and exports middleware for case conversions.

Still missing how-to doc.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/docs/decisions/0006-middleware-support-for-http-clients.rst
+++ b/docs/decisions/0006-middleware-support-for-http-clients.rst
@@ -1,0 +1,44 @@
+Middleware Support for HTTP clients
+===================================
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+We currently expose HTTP clients(axios instances) via ``getAuthenticatedHttpClient`` and ``getHttpClient`` used to make API requests
+in our MFEs. There are instances where it would be helpful if consumers could apply middleware to these clients.
+For example the `axios-case-converter <https://www.npmjs.com/package/axios-case-converter>`_ package provides
+a middleware that handles snake-cased <-> camelCase conversions via axios interceptors. This middleware would allow our MFEs to
+avoid having to do this conversion manually.
+
+Decision
+--------
+
+The ``initialize`` function provided in the initialize module initializes the ``AxiosJwtAuthService`` provided by ``@edx/frontend-platform``.
+We will add an optional param ``authMiddleware``, an array of middleware functions that will be applied to all http clients in
+the ``AxiosJwtAuthService``.
+
+Consumers will install the middleware they want to use and provide it to ``initialize``::
+
+    initialize({
+        messages: [appMessages],
+        requireAuthenticatedUser: true,
+        hydrateAuthenticatedUser: true,
+        authMiddleware: [axiosCaseConverter, (client) => axiosRetry(client, { retries: 3 })],
+    });
+
+If a consumer chooses not to use ``initialize`` and instead the ``configure`` function, the middleware can be passed in the options param::
+
+   configure({
+       loggingService: getLoggingService(),
+       config: getConfig(),
+       options: {
+            middleware: [axiosCaseConverter, (client) => axiosRetry(client, { retries: 3 })]
+       }
+    });
+
+We decided to let consumers install their own middleware packages, removing the need to install the dependency as part of ``@edx/frontend-platform``.
--- a/src/auth/AxiosJwtAuthService.js
+++ b/src/auth/AxiosJwtAuthService.js
@@ -73,6 +73,29 @@ class AxiosJwtAuthService {
         this.cachedHttpClient = this.httpClient;
         logFrontendAuthError(this.loggingService, `configureCache failed with error: ${e.message}`);
       });
+
+    this.middleware = options.middleware;
+    this.applyMiddleware(options.middleware);
+  }
+
+  /**
+   * Applies middleware to the axios instances in this service.
+   *
+   * @param {Array} middleware Middleware to apply.
+   */
+  applyMiddleware(middleware = []) {
+    const clients = [
+      this.authenticatedHttpClient, this.httpClient,
+      this.cachedAuthenticatedHttpClient, this.cachedHttpClient,
+    ];
+    try {
+      (middleware).forEach((middlewareFn) => {
+        clients.forEach((client) => client && middlewareFn(client));
+      });
+    } catch (error) {
+      logFrontendAuthError(this.loggingService, error);
+      throw error;
+    }
   }
 
   /**
@@ -89,6 +112,7 @@ class AxiosJwtAuthService {
     if (options.useCache) {
       return this.cachedAuthenticatedHttpClient;
     }
+
     return this.authenticatedHttpClient;
   }
 
@@ -104,6 +128,7 @@ class AxiosJwtAuthService {
     if (options.useCache) {
       return this.cachedHttpClient;
     }
+
     return this.httpClient;
   }
 
--- a/docs/how_tos/automatic-case-conversion.rst
+++ b/docs/how_tos/automatic-case-conversion.rst
@@ -1,0 +1,58 @@
+#####################################################################
+How to: Convert SnakeCase to CamelCase automatically for API Requests
+#####################################################################
+
+Introduction
+************
+
+When using the HTTP client from ``@edx/frontend-platform``, you are making an API request to an
+Open edX service which requires you to handle snake-cased <-> camelCase conversions manually. The manual conversion quickly gets
+tedious, and is error prone if you forget to do it.
+
+Here is how you can configure the HTTP client to automatically convert snake_case <-> camelCase for you.
+
+How do I use configure automatic case conversion?
+*************************************************
+
+You want to install `axios-case-converter <https://www.npmjs.com/package/axios-case-converter>`_, and add it
+as a middleware when calling ``initialize`` in the consumer::
+
+    import axiosCaseConverter from 'axios-case-converter';
+
+    initialize({
+        messages: [],
+        requireAuthenticatedUser: true,
+        hydrateAuthenticatedUser: true,
+        authMiddleware: [axiosCaseConverter],
+    });
+
+Or, if you choose to use ``configure`` instead::
+
+    import axiosCaseConverter from 'axios-case-converter';
+
+    configure({
+        loggingService: getLoggingService(),
+        config: getConfig(),
+        options: {
+                middleware: [axiosCaseConverter, (client) => axiosRetry(client, { retries: 3 })]
+        }
+    });
+
+By default the middleware will convert camelCase -> snake_case for payloads, and snake_case -> camelCase for responses.
+If you want to customize middleware behavior, i.e. only have responses transformed, you can configure it like this::
+    initialize({
+        messages: [],
+        requireAuthenticatedUser: true,
+        hydrateAuthenticatedUser: true,
+        authMiddleware: [(client) => axiosCaseConverter(client, {
+            // options for the middleware
+            ignoreHeaders: true, // don't convert headers
+            caseMiddleware: {
+                requestInterceptor: (config) => {
+                    return config;
+                }
+            }
+        })],
+    });
+
+See `axios-case-converter <https://github.com/mpyw/axios-case-converter>`_ for more details on configurations supported by the package.
--- a/src/auth/AxiosJwtAuthService.test.jsx
+++ b/src/auth/AxiosJwtAuthService.test.jsx
@@ -220,6 +220,36 @@ afterEach(() => {
   global.location = location;
 });
 
+describe('applyMiddleware', () => {
+  it('should apply all middleware to the http clients in the service', () => {
+    const clients = [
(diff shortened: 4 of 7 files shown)


docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
...
+
+It will look like this:
+    {
+        useCache: true,
+        middleware: [
+            axiosConvertCase,
+            [axiosRetry, { retries: 3 }],
+            ...
+        ]
+    }
```
adamstankiewicz, line 34: Let's put this in a code block so the formatting is legible.

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
...
+
+We will export any middleware functions that we support, i.e. ``applyCaseMiddleware`` from frontend-platform.
```
adamstankiewicz, line 36: I believe one of the benefits of this middleware approach is that consumers could write their own middleware functions. `@edx/frontend-platform` might export official middleware as pass-thrus (e.g., `applyCaseMiddleware`)  but theoretically consumers could use others or create their own outside of ones exported by `@edx/frontend-platform`.

In this sense, is the "... that we support..." clause accurate?

package.json
```diff
@@ -55,6 +55,7 @@
     "@cospired/i18n-iso-languages": "2.2.0",
     "axios": "0.26.1",
     "axios-cache-adapter": "2.7.3",
+    "axios-case-converter": "0.9.0",
```
adamstankiewicz, line 58: [curious] Should we be concerned about installing `axios-case-converter` if the consumer doesn't intend to use it? Maybe we should treat it more like an [optional dependency](https://betterprogramming.pub/what-are-npms-optional-dependencies-and-when-should-we-use-them-796a6a964e73)?

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
+middleware for HTTP client
+==========================================
+
+Status
+------
+
+
```
adamstankiewicz, line 7: nit: `Proposed` for now

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
...
+
+Context
+-------
+
+We currently expose a HTTP client(axios instance) via ``getAuthenticatedHttpClient`` and ``getHttpClient`` used in
+our MFEs to make HTTP requests. Since edX uses Django for its backend services, the MFEs have to handle snake-cased <-> camelCase
+conversions manually. The 'axios-case-converter' package provides a middleware that would handle the conversion automatically. In the
```
adamstankiewicz, line 14: nit: hyperlink to `axios-case-converter` on NPM

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
...
+        useCache: true,
+        middleware: [
+            axiosConvertCase,
+            [axiosRetry, { retries: 3 }],
+            ...
+        ]
+    }
+
+We will export any middleware functions that we support, i.e. ``applyCaseMiddleware`` from frontend-platform.
+The consumers will import the ones they want to use/add new ones and pass them in when creating a client.
```
adamstankiewicz, line 37: Since we are technically opening up the ability for consumers to create/apply their own middleware to the HTTP client, I wonder if it's worth including documentation such as "How do I create my own Axios client middleware?" as part of the README, etc. in `@edx/frontend-platform`, so consumers know this capability of extending the client is available to them.

For the purpose of this ADR, it might just be a "We will write documentation for how consumers can create their own middleware functions" for the time being.

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,40 @@
...
+Status
+------
+
+
+
+Context
+-------
+
+We currently expose a HTTP client(axios instance) via ``getAuthenticatedHttpClient`` and ``getHttpClient`` used in
+our MFEs to make HTTP requests. Since edX uses Django for its backend services, the MFEs have to handle snake-cased <-> camelCase
```
adamstankiewicz, line 13: nit: Open edX

docs/decisions/0006-middleware-support-for-http-clients.rst
adamstankiewicz, line 14: Is it worth calling out where/how the conversions take place (i.e., outgoing vs incoming data)?

long74100: I was going to write a separate how-to doc for this.

docs/decisions/0006-middleware-support-for-http-clients.rst
long74100: yeah I'll rephrase this part

docs/decisions/0006-middleware-support-for-http-clients.rst
long74100: Idk if it should be part of the README, but perhaps on a how-to doc, and just external links to all the examples out there.

package.json
long74100: [Chatted offline], we agreed that `axios-case-converter` should be a dependency since the intent is for consumers (mostly Open edX MFEs) to use it.

src/auth/AxiosJwtAuthService.js
```diff
@@ -75,6 +75,30 @@ class AxiosJwtAuthService {
...
+   *
+   */
+  applyMiddleware(client, middlewares = []) {
+    try {
+      (middlewares).forEach((middleware) => {
+        if (typeof middleware === 'function') {
+          middleware(client);
+        } else if (Array.isArray(middleware)) {
+          const [middlewareToApply, options] = middleware;
+          middlewareToApply(client, options);
```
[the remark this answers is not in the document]
long74100: I think we get too much benefit from validation here since it could be a bad function or options could be the wrong type. Ideally It should be obvious enough to the consumer if an exception is thrown, or if we have types.

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,41 @@
...
+------
+
+Proposed
+
+Context
+-------
+
+We currently expose a HTTP client(axios instance) via ``getAuthenticatedHttpClient`` and ``getHttpClient`` used in
+our MFEs to make HTTP requests. Since Open edX uses Django for its backend services, the MFEs have to handle snake-cased <-> camelCase
+conversions manually. The `axios-case-converter <https://www.npmjs.com/package/axios-case-converter>` package provides a middleware that would handle the conversion automatically. In the
```
adamstankiewicz, line 14: I believe the parsed link formatting here is needs some TLC (might be a missing `_` or something)

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,41 @@
...
+--------
+
+``getAuthenticatedHttpClient`` and ``getHttpClient`` take an ``options`` argument that includes configuration for the client.
+Only ``useCache`` is currently supported. We will introduce a new option ``middleware``, which will be a list of middleware
+that will be applied to the base axios instance. Each item in the list can either be a function or an array in the form of [function, options].
+We will apply the middleware and handle any exceptions that occur.
+
+It will look like this::
+
+    {
```
adamstankiewicz, line 28: One other thought for this code block is to put in the context of calling `getAuthenticatedHttpClient` as a consumer, so the depicted code isn't just the options object in isolation.

```jsx
const httpClient = getAuthenticatedHttpClient({
  useCache: true,
  middleware: [
    axiosConvertCase,
    [axiosRetry, { retries: 3 }],
    // ...
  ],
});

File: src/auth/AxiosJwtAuthService.js [hunk 11]
```diff
@@ -75,6 +75,30 @@ class AxiosJwtAuthService {
       });
   }
 
+  /**
+   *
+   * @param {HttpClient} client An axios instance.
+   * @param {Array|Function} middlewares The middleware to be applied.
```
adamstankiewicz, line 81: [nit/clarification] The JSDoc here says `middlewares` can be a function as well? I think the `Array|Function` bit comes from each element within the `middlewares` array, but the function arg passed to `applyMiddleware` would always be an array.

src/auth/AxiosJwtAuthService.js
```diff
@@ -86,10 +110,15 @@ class AxiosJwtAuthService {
...
   getAuthenticatedHttpClient(options = {}) {
+    let client = this.authenticatedHttpClient;
+
     if (options.useCache) {
-      return this.cachedAuthenticatedHttpClient;
+      client = this.cachedAuthenticatedHttpClient;
     }
-    return this.authenticatedHttpClient;
+
+    this.applyMiddleware(client, options.middleware);
```
davidjoy, line 119: If `getAuthenticatedHttpClient` is called once with options and once without, what have we done to the client, since it's the same instance?

Seems like we've introduced side effects into what was otherwise a pure function acting as a "getter" for our client instance.  Depending on the order of the calls, the client may or may not have middleware on it and it can't "go back" once they're added, either.  

Conceptually the client is configured at application initialization time; I'd expect us to add middleware to it there like we do with interceptors.  Alternately, I'd expect to add options to individual client calls (get, post, etc.) to say whether we wanted their results normalized.  Here in the getter feels like the wrong spot, admittedly.

davidjoy (changes requested):
I agree with the premise and think this is a great addition - I'm pretty sure, though, that adding it to the 'getter' functions for the client feels like the wrong place to add middleware.  See comment on AxiosJwtAuthService.  I'm happy to be a thought partner in this if it's helpful, though!

src/auth/AxiosJwtAuthService.js
davidjoy: Put another way, the `get*HttpClient` functions aren't creating a client, they're getting one that's already been configured.  Changing that configuration when we try to get a reference to it - which changes it for anyone else who called the function earlier with a different set of middleware options, feels like a bug/unexpected behavior.

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,45 @@
+Middleware Support for HTTP clients
+===================================
+
+Status
+------
+
+Proposed
```
adamstankiewicz, line 7: Accepted upon merge :)

adamstankiewicz (approved):
This looks good to me!

docs/decisions/0006-middleware-support-for-http-clients.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Context
+-------
+
+We currently expose HTTP clients(axios instances) via ``getAuthenticatedHttpClient`` and ``getHttpClient`` used in
+our MFEs to make API requests. Since Open edX uses Django for its backend services, the MFEs have to handle snake-cased <-> camelCase
```
davidjoy, line 13: This is a nit, but the ADR is about middleware support in general - may be worth framing this snake_case <-> camelCase thing as an example of a possible useful middleware.  As the second sentence in the context, it feels a bit out of nowhere.

src/auth/interface.js
```diff
@@ -95,6 +95,7 @@ let service;
 export function configure(AuthService, options) {
   PropTypes.checkPropTypes(optionsShape, options, 'property', 'Auth');
   service = new AuthService(options);
+  service.applyMiddleware(options.middleware);
```
davidjoy, line 98: I _think_ it makes sense to apply middleware in the constructor of the auth service rather than here.  

Or rather, I can't think of a reason to pull it out and not let the auth service handle it, since the implementation understands how to apply middleware and we're passing `options.middleware` into it anyway.  

If middleware were a thing that the service implementations weren't expected to understand, or was a feature of frontend-platform instead of a capability of the service itself, it might belong here.  But in spirit I think we can just let the implementation handle it itself.

For instance, some of our other functions in this file make use of a PubSub system that is extraneous to the core functionality of the service itself, but comes into play if you're using the whole frontend-platform framework together.  This doesn't feel like that - it feels like it belongs to the service implementation, not our framework around it.

Let me know if that doesn't make sense.  It's a bit of a nit, admittedly, but I also feel somewhat strongly. ;)

davidjoy (approved):
Had some comments, but also want to pre-emptively Approve so I'm not a blocker.

docs/decisions/0006-middleware-support-for-http-clients.rst
davidjoy: May just be worth setting to "Accepted" in the PR here so that it's right once it hits master. :D

PR #335, opened 2022-05-19
feat: MFE configuration at runtime

dcoa:
**Description:**

This draft PR serves as a discussion holder for the work of making mfes capable of being configured during runtime instead of during build.

This PR creates a new function to allow runtime configuration through an API call to the platform. 

 **How to test**
- To allow runtime configuration set a new environment variable MFE_CONFIG_API_URL in the env file and add the api url ( To test this you can use the API from this [PR](https://github.com/openedx/edx-platform/pull/30473) or use an external tool to mock the API response).
- The API should respond with a JSON with the config values: 
```json
{
  "SITE_NAME": "Test Site",
  "LOGO_URL": "https://testimage.com/logo.svg",
  "LOGO_TRADEMARK_URL": "https://testimage.com/logo.svg",
  "LOGO_WHITE_URL": "https://testimage.com/logo.svg",
  "FAVICON_URL": "https://testimage.com/favicon.ico",
  "CSRF_TOKEN_API_PATH": "/csrf/api/v1/token",
  "LANGUAGE_PREFERENCE_COOKIE_NAME": "openedx-language-preference",
  "LOGIN_URL": "http://testsite.example.com/login",
  "LOGOUT_URL": "http://testsite.example.com/logout",
   "STUDIO_BASE_URL": "http://studio.example.com",
  "MARKETING_SITE_BASE_URL": "http://testsite.example.com",
  "ORDER_HISTORY_URL": "http://testsite.example.com/orders",
  "REFRESH_ACCESS_TOKEN_ENDPOINT": "http://testsite.example.com/login_refresh",
}
```

- The initialize process should work normally.
_Note: You can combine buildtime and runtime configuration_

**Note: Related information [FWG Issue](https://github.com/openedx/frontend-wg/issues/103)**

**Merge checklist:**
  - [ ]  Consider running your code modifications in the included example app within frontend-platform. This can be done by running npm start and opening http://localhost:8080/. To test this you can use the API from this [PR](https://github.com/openedx/edx-platform/pull/30473) or use an external tool to mock the API response.
  - [ ]  Consider testing your code modifications in another local micro-frontend using local aliases configured via [the module.config.js file in frontend-build](https://github.com/edx/frontend-build#local-module-configuration-for-webpack). To test this you can use the API from this [PR](https://github.com/openedx/edx-platform/pull/30473) or use an external tool to mock the API response.
  - [ ]  Gather some consensus on the approach

**Post merge:**

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -67,6 +67,7 @@ let config = {
   LOGO_TRADEMARK_URL: process.env.LOGO_TRADEMARK_URL,
   LOGO_WHITE_URL: process.env.LOGO_WHITE_URL,
   FAVICON_URL: process.env.FAVICON_URL,
+  MFE_CONFIG_API_URL: process.env.MFE_CONFIG_API_URL,
 };
 
 /**
@@ -193,4 +194,5 @@ export function ensureConfig(keys, requester = 'unspecified application code') {
  * @property {string} LOGO_TRADEMARK_URL
  * @property {string} LOGO_WHITE_URL
  * @property {string} FAVICON_URL
+ * @property {string} MFE_CONFIG_API_URL
  */
--- a/src/config.js
+++ b/src/config.js
@@ -68,6 +68,7 @@ let config = {
   LOGO_WHITE_URL: process.env.LOGO_WHITE_URL,
   FAVICON_URL: process.env.FAVICON_URL,
   MFE_CONFIG_API_URL: process.env.MFE_CONFIG_API_URL,
+  APP_ID: process.env.APP_ID,
 };
 
 /**
@@ -195,4 +196,5 @@ export function ensureConfig(keys, requester = 'unspecified application code') {
  * @property {string} LOGO_WHITE_URL
  * @property {string} FAVICON_URL
  * @property {string} MFE_CONFIG_API_URL
+ * @property {string} APP_ID
  */
--- a/src/initialize.js
+++ b/src/initialize.js
@@ -51,7 +51,7 @@ import {
 } from './pubSub';
 // eslint-disable-next-line import/no-cycle
 import {
-  getConfig, mergeConfig, setConfig,
+  getConfig, mergeConfig,
 } from './config';
 import {
   configure as configureLogging, getLoggingService, NewRelicLoggingService, logError,
@@ -131,7 +131,6 @@ export async function auth(requireUser, hydrateUser) {
     hydrateAuthenticatedUser();
   }
 }
-
 /*
  * Set or overrides configuration through an API.
  * This method allows runtime configuration.
@@ -140,23 +139,25 @@ export async function auth(requireUser, hydrateUser) {
 
 export async function runtimeConfig() {
   try {
-    const apiConfig = { headers: { accept: 'application/json' } };
-    const apiService = await configureCache();
+    const { MFE_CONFIG_API_URL, PUBLIC_PATH } = getConfig();
 
-    const url = parseUrlQueryParams(getConfig().MFE_CONFIG_API_URL, [
-      { name: 'mfe', value: (getConfig().PUBLIC_PATH.slice(1, -1) || window.location.host.split('.')[0]) },
-    ]);
+    if (MFE_CONFIG_API_URL) {
+      const apiConfig = { headers: { accept: 'application/json' } };
+      const apiService = await configureCache();
 
-    const { data } = await apiService.get(url, apiConfig);
-    mergeConfig(data);
+      const url = parseUrlQueryParams(MFE_CONFIG_API_URL, [
+        { name: 'mfe', value: (PUBLIC_PATH.slice(1, -1) || window.location.host.split('.')[0]) },
+      ]);
+
+      const { data } = await apiService.get(url, apiConfig);
+      mergeConfig(data);
+    }
   } catch (error) {
     // eslint-disable-next-line no-console
     console.error('Error with config API', error.message);
-    setConfig({
-      BASE_URL: window.location.host,
-    });
   }
 }
+
 /**
  * The default handler for the initialization lifecycle's `analytics` phase.
  *
@@ -252,7 +253,8 @@ export async function initialize({
     publish(APP_PUBSUB_INITIALIZED);
 
     // Configuration
-    if (getConfig().MFE_CONFIG_API_URL) { await runtimeConfig(); } else { await handlers.config(); }
+    await handlers.config();
+    await runtimeConfig();
     publish(APP_CONFIG_INITIALIZED);
 
     // Logging
--- a/src/initialize.js
+++ b/src/initialize.js
@@ -138,14 +138,14 @@ export async function auth(requireUser, hydrateUser) {
 
 export async function runtimeConfig() {
   try {
-    const { MFE_CONFIG_API_URL, PUBLIC_PATH } = getConfig();
+    const { MFE_CONFIG_API_URL, APP_ID } = getConfig();
 
     if (MFE_CONFIG_API_URL) {
       const apiConfig = { headers: { accept: 'application/json' } };
       const apiService = await configureCache();
 
       const params = new URLSearchParams();
-      params.append('mfe', (PUBLIC_PATH.slice(1, -1) || window.location.host.split('.')[0]));
+      params.append('mfe', APP_ID);
       const url = `${MFE_CONFIG_API_URL}?${params.toString()}`;
 
       const { data } = await apiService.get(url, apiConfig);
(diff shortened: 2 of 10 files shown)


openedx-webhooks:
Thanks for the pull request, @dcoa! I've created [OSPR-6698](https://openedx.atlassian.net/browse/OSPR-6698) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

natabene:
@dcoa Thank you for your contribution.

dcoa:
Hi @natabene  this PR is ready for review.

MaferMazu:
I think this should be on your radar @arbrandes

dcoa:
@natabene could you help me re-run the test, please?

dcoa:
Hi, @davidjoy  @adamstankiewicz  I'm working on changes to make it possible to read the response of the API to configure frontend applications at runtime, I hope you can check this pr and give us feedback 🙏

I have this https://github.com/openedx/frontend-app-account/pull/603 too to make it possible to change the favicon and the site name in the title tag, could you check it?

src/initialize.js
```diff
@@ -223,7 +252,7 @@ export async function initialize({
     publish(APP_PUBSUB_INITIALIZED);
 
     // Configuration
-    await handlers.config();
+    if (getConfig().MFE_CONFIG_API_URL) { await runtimeConfig(); } else { await handlers.config(); }
```
davidjoy, line 255: I think I'd suggest moving this `if` statement logic _into_ the default config handler.  This logic here would continue to just say `await handlers.config();` and we would optionally call the `runtimeConfig` function if `MFE_CONFIG_API_URL` exists.  That way we enable folks to continue to override all of our default config handling.

Note that currently the `config` handler is set to `noOp`, so we'll just need to create a new one similar to `auth`, `analytics`, or `initError`.  Let me know if this doesn't make sense!

src/initialize.test.js
```diff
@@ -240,4 +278,83 @@ describe('initialize', () => {
     expect(overrideHandlers.ready).not.toHaveBeenCalled();
     expect(overrideHandlers.initError).toHaveBeenCalledWith(new Error('uhoh!'));
   });
+
+  it('should initialze the app with runtime configuration', async () => {
```
davidjoy, line 282: Nit: `initialze` -> `initialize`

davidjoy (commented):
Still looking and don't have a lot of time today, but wanted to give some early feedback!  Will be able to look more closely on Monday.

src/initialize.js
dcoa: I consider this, but many applications have the override handler to marge additional values for that specific MFE ([account example](https://github.com/openedx/frontend-app-account/blob/c39fd332b670f92879db3d227aaba83a1ad44e26/src/index.jsx#L60)), which make even if you have MFE_CONFIG_API_URL don't call the API.

That's why I put it there because you can set that additional values with the API or set by env file.

May be I could move the _if_ into `runtimeConfig` and call like: 
```javascript
    await handlers.config(); 
    await runtimeConfig(); 
```

What do you think?

src/initialize.js
```diff
@@ -128,6 +132,31 @@ export async function auth(requireUser, hydrateUser) {
...
+    const url = parseUrlQueryParams(getConfig().MFE_CONFIG_API_URL, [
+      { name: 'mfe', value: (getConfig().PUBLIC_PATH.slice(1, -1) || window.location.host.split('.')[0]) },
+    ]);
+
+    const { data } = await apiService.get(url, apiConfig);
+    mergeConfig(data);
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Error with config API', error.message);
+    setConfig({
```
davidjoy, line 155: I _think_ we should let the MFE keep its build-time configuration rather than using `setConfig` here.  I think it'd be more correct to leave all the build-time config values in there, rather than removing everything but BASE_URL.  The way this is now, all the values will start returning `undefined` instead of whatever's in the build-time config, which will likely be reasonable fallbacks in many cases.

src/config.js
```diff
@@ -67,6 +67,7 @@ let config = {
   LOGO_TRADEMARK_URL: process.env.LOGO_TRADEMARK_URL,
   LOGO_WHITE_URL: process.env.LOGO_WHITE_URL,
   FAVICON_URL: process.env.FAVICON_URL,
+  MFE_CONFIG_API_URL: process.env.MFE_CONFIG_API_URL,
```
davidjoy, line 70: I think this value needs to be added in a few more places.  As a way to discover places it should appear, I tried searching for `LOGO_TRADEMARK_URL` above and noted that config values in this file are also set in:

- `.env.test`
- `.env.development` 
- `setupTest.js`

Basically, it needs to be defaulted in those files for completeness sake.

src/initialize.js
```diff
@@ -127,6 +131,32 @@ export async function auth(requireUser, hydrateUser) {
...
+export async function runtimeConfig() {
+  try {
+    const { MFE_CONFIG_API_URL, PUBLIC_PATH } = getConfig();
+
+    if (MFE_CONFIG_API_URL) {
+      const apiConfig = { headers: { accept: 'application/json' } };
+      const apiService = await configureCache();
+
+      const url = parseUrlQueryParams(MFE_CONFIG_API_URL, [
+        { name: 'mfe', value: (PUBLIC_PATH.slice(1, -1) || window.location.host.split('.')[0]) },
```
davidjoy, line 149: Hmm... I don't know if we can assume that the MFE name exists in the URL. 😬  I didn't notice this earlier.  While we use the PUBLIC_PATH to know where the MFE is served from, I don't think we can otherwise make any assumptions about what the URL is or what information it might contain.  Today an MFE can be deployed to any sub-domain or path and it doesnt' matter, it's self-contained.

I'm thinking about the use cases for runtime config now.  If we want to use it to change styling/naming at runtime because we want to support multi-tenancy on a single MFE deployment, do we expect our edx-platform configuration to have different MFE config for each tenant?  Or do we want to use runtime config to simply replace build-time configuration?

I think I'd rather see us add an `APP_ID` environment variable to MFEs to specify the identifier of an MFE for use cases like this where a service needs to know which MFE is calling, but I don't think that helps the multi-tenancy use case.  

What use cases does eduNEXT need to support here / how do we expect to configure an MFE for multi-tenancy, which I assume is a concern?

src/utils.js
```diff
@@ -165,3 +165,17 @@ export function ensureDefinedConfig(object, requester) {
...
+
+/**
+ * This function helps to parse api query params
+ *
+ * @param {String} mainURI
+ * @param {Array} queryParams Array with object with strings to set query name and value
+ *  {name: '', value, ''}
+ */
+
+export function parseUrlQueryParams(mainURI, queryParams = []) {
```
davidjoy, line 177: I'm sensitive to us adding more utility functions here.  If we add something to the utilities here, it should be something we expect MFEs will need; these are intended to be common operations that MFEs will generally all do.  Since `parseUrlQueryParams` is just used internally, I think I'd almost rather we just inline it where we need it:

```
const params = new URLSearchParams();
params.append('mfe', <mfe name>);
const url = `${MFE_CONFIG_API_URL}?${params.toString()}`;
```

Having those three lines up in the `runtimeConfig` function feels pretty clear and simple to understand.

src/initialize.js
dcoa: This use case is set or overrides the MFE configuration at runtime, not covering the multitenancy case.

This is a proposal [roadmap](https://docs.google.com/document/d/1-FHIQmyeQZu3311x8eYUNMru4JX7Yb3UlqjmJxvM8do/edit#heading=h.m3u89l2084ecl) you can comment and discuss it.

src/config.js
dcoa: I could add the key without value by default because is optional  :)

src/initialize.js
dcoa: I add de `APP_ID`

arbrandes: For the record, I think `APP_ID` is a great idea. +1.

dcoa:
Hi @davidjoy  @arbrandes, I see the test failed and find that is an issue with npm you can check a better explanation here https://github.com/npm/cli/issues/4998 

I see this [PR](https://github.com/openedx/frontend-platform/pull/345) solved it by pinning the npm version, should I do the same here or wait for the merge of that to re-run the test?

Thanks for your help.

davidjoy:
I think we can get that other PR merged pretty quick and available on master for this branch to rebase onto.  I think if you made the same change here we might just end up with a conflict instead. :)

dcoa:
Great, I wait for the npm pinning version to rebase mi branch, thank you :)

dcoa:
Hi, @davidjoy @arbrandes, I already rebase the branch, could you help me re-run the test, please  😀

dcoa:
Yes, all the tests are passing with the latest changes

arbrandes (approved):
Looks good to me!  Once again, great work @dcoa.

davidjoy (approved):

davidjoy:
Great work, @dcoa!  And thank you for being so flexible, I know I nit-picked the details. 😄

dcoa:
@davidjoy @arbrandes @felipemontoya 

Thanks a lot for your help and feedback.

Could you help me to merge it? 😊

PR #342, opened 2022-06-06
feat: export useIntl and createIntl from react-intl

muselesscreator:
**Description:**

Exports useIntl and createIntl from react-intl library.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/src/i18n/index.js
+++ b/src/i18n/index.js
@@ -72,6 +72,7 @@
  */
 
 export {
+  createIntl,
   FormattedDate,
   FormattedTime,
   FormattedRelativeTime,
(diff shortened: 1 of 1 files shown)


src/i18n/index.js
```diff
@@ -72,6 +72,8 @@
  */
 
 export {
+  createIntl,
+  intlShape,
```
adamstankiewicz, line 76: `intlShape` is already exported below from `'./lib`. In v3 of `react-intl`, `intlShape` was removed from the library so as part of the upgrade to v5, we provided own our copy of `intlShape` instead below so it doesn't break for consumers.

muselesscreator: noted!

PR #351, opened 2022-07-06
docs: ADR that proposes a common stylesheet for all MFEs

xitij2000:
**Description:**

Adds an ADR that proposes creating a single common stylesheet that all MFEs load from a central place.

**Merge checklist:**

**Post merge:**


docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,116 @@
...
+separate stylesheet and built separately from the MFE-specific stylesheet.
+
+Both these stylesheets will be included in each MFE, however the common
+stylesheet can be loaded from a common source for all MFEs.This will allowing
+for deploying the common theme stylesheet without redeploying the MFEs.
+
+The mechanism for loading this common stylesheet will need to be different from
+the mechanism for loading the MFE stylesheet. A JavaScript theme loader will
+contain the location of the SCSS file and will add this stylesheet to the
+document.
```
xitij2000, line 55: The exact mechanism of this is currently not in this ADR but will be the subject of another ADR if this is accepted. 

The prototype presented [here](https://github.com/openedx/frontend-app-discussions/pull/125) uses Webpack's module federation, however for the purpose of just a single stylesheet, this may be an overkill. It might make sense to create a specific loader for this.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,116 @@
...
+Rejected Alternatives
+*********************
+
+- Loading the common CSS directly instead of with a JavaScript based loader
+  Adding a loader eliminates caching issues. Currently the CSS files include
+  the hash of the CSS code in the file name, so that with each redeployment the
+  file names change, ensuring that a file is cached as long as it's used but not
+  loaded any more once the file contents change.
+
+  Since the common theme needs to have a fixed name, we can't rely on changing
```
bradenmacdonald, line 95: This confused me at first. I think what you're saying is "Without using a JavaScript based loader on the client side, and without support for pre-rendering our MFEs on the server side, we would need to use a fixed URL/filename for the common CSS file, which would cause problems with caching that file."

Because in your actual ADR, the common theme does not need to have a fixed name ("now it can be a file with
  the hash in its name").

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,116 @@
...
+from there. Otherwise, it can continue using the existing mechanism of building
+and deploying the theme for each MFE for backwards-compatibility. It can also
+provide hooks for loading the theme and potentially/eventually also selecting
+a theme.
+
+Consequences
+************
+
+With a single common stylesheet for all MFE, they should load quicker since the
+stylesheet can be cached once for all MFEs. It also makes theme rebuilds much
```
bradenmacdonald, line 74: I understand the theming benefits listed in the second half of this paragraph.
What I'm wondering about is the first benefit, loading the MFEs quicker since the stylesheet is cached. Could this also be achieved just by deploying all MFE's CSS assets to the same CDN folder? Because if those four lines (fonts, variables, core, overrides) are the same for each MFE, and the same version of Paragon and theme are used by each MFE, and webpack is configured correctly, it should produce a CSS chunk file for those core styles with a hash in its name, and that hash should be the same for each such MFE. So by deploying the CSS to the same CDN location, the different MFEs would still load the same common CSS file without any other changes. Or no?

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,116 @@
...
+quicker, and eliminates the need for rebuilding each MFE for minor change to
+the theme.
```
bradenmacdonald, line 76: For context, roughly how long does a "theme rebuild" for all MFEs take today, and how long would it take with this change?

bradenmacdonald (commented):
As I understand it, one of the big benefits of MFEs is the ability to very quickly iterate on an individual MFE and deploy changes. Can you talk about that in the context of this ADR? e.g. does this ADR require that all MFEs are using the same version of Paragon, or is it still possible for one MFE to be using a newer version of Paragon (and newer common CSS) while other MFEs haven't yet been updated (and so use a different common CSS file?). And in that case, if multiple MFEs can use different versions of the common CSS file, are theme updates still faster?

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
xitij2000: I think this will require making specific assumptions about how it's built and loaded from a CDN. Additionally, as mentioned elsewhere, currently different MFEs can use a slightly different version of Paragon, which will cause the hashes to not line up.

xitij2000:
> As I understand it, one of the big benefits of MFEs is the ability to very quickly iterate on an individual MFE and deploy changes. Can you talk about that in the context of this ADR? e.g. does this ADR require that all MFEs are using the same version of Paragon, or is it still possible for one MFE to be using a newer version of Paragon (and newer common CSS) while other MFEs haven't yet been updated (and so use a different common CSS file?). And in that case, if multiple MFEs can use different versions of the common CSS file, are theme updates still faster?

I've experimented with this a bit, and in my experience things work pretty well even with different versions of Paragon across MFEs. That said, there can be sources if issues, one being the following situation:

There is a bug in a Paragon component that requires a JavaScript and CSS fix. The version of Paragon including the bug fix is updated for a particular MFE, but not for the common theme. So the JavaScript for the Paragon bug fix does get deployed, but not the CSS code. 

I think it will be important to keep Paragon in the common theme pinned to the latest version used by all MFEs, and to redeploy the common theme each time an MFE is updated and redeployed if it includes an update to Paragon.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
bradenmacdonald: Cool... I think it's worth incorporating that example into the text of the ADR.

bradenmacdonald:
> There is a bug in a Paragon component that requires a JavaScript and CSS fix. The version of Paragon including the bug fix is updated for a particular MFE, but not for the common theme. So the JavaScript for the Paragon bug fix does get deployed, but not the CSS code.

Hmm, that does seem like a significant drawback. Could it make sense to include the JS and CSS together in the common theme code, so that they're always in sync at least?

Also: It seems that the main benefits of this ADR accrue when someone is actively editing or updating themes. But I imagine for each given Open edX instance, that will only occur occasionally (e.g. when first setting the instance up), and 99% of the time it will be operated and maintained without anyone making changes to theming. (Or is that a wrong assumption?). If so, would it make sense to have a kind of "theming mode" where this ADR's functionality is enabled, but otherwise leave it off for normal operations and faster deployment of MFE updates? Or alternatively, would it make sense to use the approach specified in this ADR for instance using named release versions (favoring theming flexibility), but use the current method for instances that are deployed continuously (favoring flexibility, accuracy, and rapid iteration for each individual MFE)?

bradenmacdonald:
Cool. Sorry if I muddied the waters :p

natabene:
@xitij2000 Thank you for your contribution. Please let me know once this is ready for a review.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,139 @@
...
+making the theme loader part of the init process so it runs before the app is
+considered ready.
+
+Another major consequence of this change would be that the CSS code for Paragon
+would be loaded using a different mechanism than the JavaScript code of Paragon.
+The CSS code would be compiled into the theme whereas the JS code will be loaded
+directly. This means that if different versions of Paragon are used for an MFE
+vs the common theme, then there is a chance that an incompatible version of the
+CSS could be loaded for the MFE. This will make it all the more important to
+ensure that version of Paragon used across all MFEs is compatible. This will be
```
felipemontoya, line 96: I am imagining that making paragon compatible only happens one time. Before a mfe uses this compatible version of paragon then it is business as usual. After, they are loading the common css and that is it.

Do you foresee that there could be other moments where compatibility is broken?

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,139 @@
...
+the theme.
+
+For context, on a sample development machine, rebuilding an MFE takes roughly
+one minute, of this, on the same machine, around 3-4 seconds is the time it
+takes to build the theme.
+
+By continuing to support the existing theming mechanism, the current mechanism
+for theming can continue working for those who need it.
+
+Since there is a JavaScript-based loader involved, there is potential for
```
felipemontoya, line 85: I think that getting it right for the loader is the gist of this proposal.

Would it be possible that there is more than a loader written and which loader is used is also a configuration? I am looking at this from the lens of multitenancy probably requiring a more complex/slower loader which not everyone should pay the penalty for.

felipemontoya (approved):
I like the proposal of this ADR. As always, the detail of the implementation will be the critical factor but the proposed road looks good to me.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
```diff
@@ -0,0 +1,139 @@
...
+to all of them, which is the following four lines:
+
+.. code-block:: SCSS
+
+	@import "~@edx/brand/paragon/fonts";
+	@import "~@edx/brand/paragon/variables";
+	@import "~@edx/paragon/scss/core/core";
+	@import "~@edx/brand/paragon/overrides";
+
+These four lines, import the Paragon and bootstrap code and apply
```
dcoa, line 23: Hi @xitij2000, I think this is a good idea with the actual status of the theming. 

I was wondering about the header, footer, and cookie banner styles, that should be updated to use CSS variables, update the version in the MFE and be part of the MFE stylesheet or could be part of the theme stylesheet?

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
xitij2000: I'm not sure I understood what you meant by "making paragon compatible only happens one time". 

I think the issue I'm looking at is as follows:
- A bug is encountered in a Paragon component. For example if a label/text for some component is longer than a certain threshold, it will break the UI. 
- The fix for this involves some React code and a change in SCSS. 
- The common compiled theme is built on an older version of Paragon that doesn't include this fix. 
- An MFE is updated to include a newer version of Paragon. 
- The MFE is now using React code, but it's supporting CSS code isn't available in the theme. 

This would need the common theme to be rebuilt as well, using the latest Paragon.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
xitij2000: In this proposal I imagine that the loader can be as simple as:

```JavaScript
const urlOfThemeCSS= '...';

addCSSStyleSheet(urlOfThemeCSS);
```
Where the function will essentially create a `link` element to load the stylesheet. In addition, it can have a simple API to signal to the app when the theme has loaded. 

I think a good path for multi-tenancy would be to combine this with the great work being done on runtime configuration: https://github.com/openedx/frontend-wg/issues/103

Since the loader path is a config variable, and the runtime configuration can provider URLs that vary based on the site, it would allow pointing to different themes. The loader itself can just remain simple, if that makes sense.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
xitij2000: I think the header, footer, cookie banner and other elements shoudl also over time be updated to use CSS variables. They can then be independent stylesheets that are loaded for those. 

So we'd have:

1. Base theme stylesheet (loaded from a common source)
2. Component stylesheets (header, footer, banner etc. separate for each MFE)
3. MFE stylesheet (separate for each MFE).

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
felipemontoya: I understand now. I thought you were the initial compatibility change, but this issue will come about constantly.

docs/decisions/0007-compile-common-theme-for-all-mfes.rst
felipemontoya: What you are saying then is that the loader sis so simple that there will be no need for more than one. If the loader ends up being something so flexible, then the config variable will take care of everything.

I was trying to avoid a situation where there is already logic baked into the loader that makes it imposible to extend via configuration. It does not look like that is the case though.

xitij2000: The loader ideally **needs** to be simple, since the idea is that it won't be cached. So the smaller it is the better. 

Right now the scss name is baked into the index.html file. If the stylesheet needs to load dynamically we can't do that. So we need some way to dynamically determine the name of this stylesheet file. One mechanism is to have a more complex API that returns the stylesheet location. An alternative is to have a small javascript file (the loader) that has the location of the stylesheet in it. Whenever the stylesheet location changes this file will be updated, and it will always be reloaded, so we get the latest stylesheet. 

So making the loader heavy would be detrimental to the loading speed, though there is probably nothing stopping people from doing it.

felipemontoya: So in the html we would have:
```html
<!DOCTYPE html>
<html>
<head>
    <script type="text/javascript" src="loaderfile_with_no_cache.js"></script>
    <script type="text/javascript" src="long_react_code.123HASH134.js"></script>
</head>
```
And if someone wants to override the loader, they can go and configure the thing to load the file from anywhere else?

e.g:
```
<!DOCTYPE html>
<html>
<head>
    <script type="text/javascript" src="different/path/custom_loaderfile_with_no_cache.js"></script>
    <script type="text/javascript" src="long_react_code.123HASH134.js"></script>
</head>
```

xitij2000: @felipemontoya I think we'd load the loader itself dynamically, and just append the current timestamp as a query parameter to break the cache. 

However, the location of this loader would be a config variable provided during build, or with the runtime configuration mechanism dynamically during runtime.

mphilbrick211:
@arbrandes have you had a chance to review this?

arbrandes:
@mphilbrick211, as far as I understand, this is actually blocked on a separate discussion on [Design Tokens](https://github.com/openedx/paragon/issues/1546), and maybe other things as well.  @xitij2000 may be able to tell us if this will require changes.

mphilbrick211:
Hi @xitij2000! Just following up on Adolfo's comment. Is this still blocked?

xitij2000:
> Hi @xitij2000! Just following up on Adolfo's comment. Is this still blocked?

Sorry for losing track of this. I will look into the current status of design token and how that would impact this ADR over the next few days.

xitij2000:
I've had a look at how the design tokens work will impact this, and I feel that while the basic idea remains the same, i.e. a single core stylesheet for each MFE, some of the mechanisms here will change on that basis. 

A bigger impact though is bound to come from the [modular MFE domains work](https://github.com/openedx/open-edx-proposals/pull/410). Since that work will likely involve consolidating MFEs into single container MFE and at that point, having a single instance of the theme shared across MFEs will be an easier reality, especially if they are all sharing the same version of Paragon. 

I've done some experiments on this and, with a container SPA hosting the theme the need for this ADR may go away. I'd like to keep this around for now till those projects are in a more defined place, so I can either close or modify this.

adamstankiewicz:
> with a container SPA hosting the theme the need for this ADR may go away.

@xitij2000 I agree, it definitely makes sense that any child MFEs of the container SPA would inherit from the theme applied at the container SPA layer. That said, during local development, I believe we may want to strive for being able to run child MFEs independent of the container SPA (if possible), e.g. if I'm not making changes to the container SPA, my dev environment ideally shouldn't need to run it.

This is kind of similar to when `@edx/paragon` is installed in a JS library as a peer dependency but also installed as a dev dependency for local development in the library.

[curious] In that case, I'd wonder if this ADR might potentially still be relevant?

This is also an area we'd love to get your input on as it relates to the Paragon design tokens project. For example, the `@edx/brand-edx.org` repo with design tokens generates a CSS file itself containing the CSS variables based on the customized design tokens. This CSS file could then be hosted on a CDN and then largely referenced by consuming applications. For example, perhaps child MFEs rely on the container SPA in production, but import the hosted CSS theme file during development.

xitij2000:
> That said, during local development, I believe we may want to strive for being able to run child MFEs independent of the container SPA (if possible), e.g. if I'm not making changes to the container SPA, my dev environment ideally shouldn't need to run it (or at least a local copy of it).

I think this will depend a lot on the mechanism being used for this. I was experimenting on an approach using a monorepo for all learner MFEs using npm workspaces. i.e. each MFE is a package inside a larger global repo, with MFEs loading each other using module federation. This, if set up properly would allow using each MFE individually, but also via the container SPA. 

If the mechanism that's finally settled on is similar, then it might still be possible to have the theme installed and built for the container SPA but still use and reference it in individual MFEs. 

For other approaches, I agree, the approach for individual MFEs in local development can remain much the same, i.e. build a local copy of the branding package point the individual MFE to it.

mphilbrick211:
Hi @xitij2000 and @arbrandes just checking in on this :)

arbrandes:
@mphilbrick211, @xitij2000 might have more to say, but I believe this is in the backburner, pending further development on https://github.com/openedx/open-edx-proposals/pull/410 and https://github.com/openedx/paragon/issues/1546.  Both will have a significant impact on how theming will be done in the future.

adamstankiewicz:
[inform] @viktorrusakov and I put together a prototype of what we imagine could work to dynamically load common stylesheet(s) generated by Paragon's design tokens into MFEs might look like: https://github.com/openedx/frontend-platform/pull/440

Would love any feedback on the proposed approach in the above PR.

tl;dr; 
* it relies on 2 new config variables being present (3 in the future):
  * `APP_THEME_CORE_URL` (core theme CSS)
  * `APP_THEME_LIGHT_URL` (CSS variable definitions specific to light theme variant)
  * [in the future] `APP_THEME_DARK_URL` (CSS variables definitions specific to dark theme variant, one day)
* Implements `useAppTheme` React hook to inject the external theme URLs into the HTML `<head>`.
* Does not render MFE until the external CSS is loaded to avoid Flash of Unstyled Content (FoUC).
* Ensures that once we do support more than 1 theme variant (i.e., both light _and_ dark theme variants), switching between them does not introduce FoUS.
* Provides a mechanism for consuming MFEs to (eventually) switch the theme at runtime by exposing a dispatch function and a set of reducer actions to mutate the app theme state (e.g., a consuming MFE could do `setAppThemeVariant('dark')` in the future once more theme variants beyond "light" is supported.

cc @xitij2000 @arbrandes

mphilbrick211:
@arbrandes @adamstankiewicz just checking in on this to see if there are any updates.

xitij2000:
> @arbrandes @adamstankiewicz just checking in on this to see if there are any updates.

I think the new PR linked by @adamstankiewicz above makes it possible to implement this in practice, however I'm still not up to date on the modular MFE discovery and whether it resolves the issue of different MFEs using different version of Paragon. 

With that PR some of the reasons for using a JavaScript loader are no longer applicable. Since the variables.css file is likely to be much smaller, it might be OK for it to have a short caching period, so it can be quickly updated. Or, with each variable a random query param can be added to the config API response so that the URL is forcefully refreshed.

arbrandes:
@xitij2000 

> (...) not up to date on the modular MFE discovery and whether it resolves the issue of different MFEs using different version of Paragon.

That's one of the first goals to solve, but we don't have the details, yet.  But it's heading towards a world where MFEs will be encouraged to use Paragon from the containing SPA, but we won't  be able to _force_ them to do it.  (This entails Paragon being very careful about backwards compatibility, but we knew that from the get-go.)

Given that not using a global Paragon will likely exclude the MFE from features such as the one proposed in this PR, I expect all of them - or at least, all of the ones where this feature would be important - to follow the same pattern.

xitij2000:
> @xitij2000
> 
> > (...) not up to date on the modular MFE discovery and whether it resolves the issue of different MFEs using different version of Paragon.
> 
> That's one of the first goals to solve, but we don't have the details, yet. But it's heading towards a world where MFEs will be encouraged to use Paragon from the containing SPA, but we won't be able to _force_ them to do it. (This entails Paragon being very careful about backwards compatibility, but we knew that from the get-go.)
> 
> Given that not using a global Paragon will likely exclude the MFE from features such as the one proposed in this PR, I expect all of them - or at least, all of the ones where this feature would be important - to follow the same pattern.

Do you think it makes sense to update this discovery to take into account the recent developments with style dictionary, and @adamstankiewicz's PR: https://github.com/openedx/frontend-platform/pull/440, given that the goal of modular MFEs is already to align all MFEs to the same version of paragon?

I'm planning to go through the modular MFE discovery soon to familiarise myself with it as well.

arbrandes:
@xitij2000,

> Do you think it makes sense to update this discovery to take into account the recent developments with style dictionary

We're very much taking it into account, yes! :)

mphilbrick211:
Hi @arbrandes and @xitij2000 - just following up to see if there's an update here?

xitij2000:
> Hi @arbrandes and @xitij2000 - just following up to see if there's an update here?

I'm going to spend some time updating this PR this week so it aligns better with the latest developments.

xitij2000:
> Hi @arbrandes and @xitij2000 - just following up to see if there's an update here?

I've updated the PR to more closely match the current implementation.

mphilbrick211:
Just checking in on this @xitij2000 :)

xitij2000:
> Just checking in on this @xitij2000 :)

I had updated it to match the new implementation as of that comment. If there have been further changes I will allocate some time next week to update it again.

mphilbrick211:
Hi @xitij2000 and @arbrandes! Just checking in on this :)

xitij2000:
> Hi @xitij2000 and @arbrandes! Just checking in on this :)

I've updated the ADR to match changes, however I think it might be better to wait till that work is over, and have @adamstankiewicz  work on this doc since I'm no longer fully up to date on the nuances of the recent changes.

mphilbrick211:
> > Hi @xitij2000 and @arbrandes! Just checking in on this :)
> 
> I've updated the ADR to match changes, however I think it might be better to wait till that work is over, and have @adamstankiewicz work on this doc since I'm no longer fully up to date on the nuances of the recent changes.

Hi @xitij2000 @adamstankiewicz - checking back in on this :)

xitij2000:
> > > Hi @xitij2000 and @arbrandes! Just checking in on this :)
> > 
> > 
> > I've updated the ADR to match changes, however I think it might be better to wait till that work is over, and have @adamstankiewicz work on this doc since I'm no longer fully up to date on the nuances of the recent changes.
> 
> Hi @xitij2000 @adamstankiewicz - checking back in on this :)

Sorry I haven't had much time to check into this due to time off recently. I don't think the PR this is tracking has changed since I last updated this ADR.

Mashal-m:
Hey @xitij2000 , What is the current status of this PR, is it ready to review and merge?

xitij2000:
I am going to close this PR for now and look into it when the paragon PRs are closer to final.

openedx-webhooks:
@xitij2000 Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-platform&entry.867055334=https://github.com/openedx/frontend-platform/pull/351&entry.1484655318=https://github.com/xitij2000&entry.752974735=2022-07-06+05:50&entry.1917517419=2023-11-20+12:35&entry.2133058324=No) so we can improve your experience in the future.

openedx-webhooks:
@xitij2000 Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-platform&entry.867055334=https://github.com/openedx/frontend-platform/pull/351&entry.1484655318=https://github.com/xitij2000&entry.752974735=2022-07-06+05:50&entry.1917517419=2023-11-20+12:35&entry.2133058324=No) so we can improve your experience in the future.

PR #354, opened 2022-08-05
Updated front-build to v12

BilalQamar95:
**Ticket:**
[42: Upgrade eslint to v8.x](https://github.com/openedx/frontend-wg/issues/42)

**Description**
- Updated `frontend-build `to v12 (Eslint was updated in `frontend-build` version resulting in it's version being bumped to v12. This PR updates `frontend-build` to reciprocate eslint version update)
- Resolved `eslint` issues post update

**Merge checklist:**

**Post merge:**

### DIFF
--- a/src/react/AppProvider.jsx
+++ b/src/react/AppProvider.jsx
@@ -61,13 +61,13 @@ export default function AppProvider({ store, children }) {
     setLocale(getLocale());
   });
 
+  const appContextValue = useMemo(() => ({ authenticatedUser, config, locale }), [authenticatedUser, config, locale]);
+
   return (
     <IntlProvider locale={locale} messages={getMessages()}>
       <ErrorBoundary>
         <AppContext.Provider
-          value={
-            useMemo(() => ({ authenticatedUser, config, locale }), [authenticatedUser, config, locale])
-          }
+          value={appContextValue}
         >
           <OptionalReduxProvider store={store}>
             <Router history={history}>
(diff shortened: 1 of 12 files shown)


src/react/AppProvider.jsx
```diff
@@ -65,11 +65,13 @@ export default function AppProvider({ store, children }) {
     <IntlProvider locale={locale} messages={getMessages()}>
       <ErrorBoundary>
         <AppContext.Provider
-          value={{ authenticatedUser, config, locale }}
+          value={
+            useMemo(() => ({ authenticatedUser, config, locale }), [authenticatedUser, config, locale])
```
davidjoy, line 69: As a point of style, I'd prefer to remove this `useMemo` hook from the JSX and create an intermediate variable that we then pass into the `value` attribute here.  

```
const appContextValue = useMemo(() => ({ authenticatedUser, config, locale }), [authenticatedUser, config, locale]);
```

... That sorta thing.

src/react/PageRoute.jsx
```diff
@@ -22,7 +22,7 @@ export default function PageRoute(props) {
     if (match) {
       sendPageEvent();
     }
-  }, [JSON.stringify(match)]);
```
davidjoy, line 25: This change makes me nervous - are we sure that the `match` object changes with the same frequency as the stringified version of it?  I'd be worried that we end up sending page events far more frequently, which could be pretty disruptive to our analytics.  If this was just to satisfy the linter I'd probably revert this and leave it as-is.  We should try not to make code changes when just updating the build library.

davidjoy (changes requested):
I think this commit should be a `fix:` since we're touching code that runs in our apps - it's not just a `chore` update of the build library.

See comments inline, there's one thing I think I'd factor a bit differently, and another which I'd hesitate to include in this PR.

davidjoy (approved):

PR #393, opened 2022-11-02
doc: update document for major release guide

leangseu-edx:
**Description:**

Adding PR template for upgrading other package when major release happen.

**Merge checklist:**

**For Major Release Only**
- [ ] Update dependency/peer dependency repo to support the breaking change
  - [ ] https://github.com/openedx/frontend-component-footer
    - Example: https://github.com/openedx/frontend-component-footer/pull/241
  - [ ] https://github.com/openedx/frontend-component-header
    - Example: https://github.com/openedx/frontend-component-header/pull/275
  - [ ] https://github.com/edx/frontend-component-footer-edx
    - Example: https://github.com/edx/frontend-component-footer-edx/pull/208
  - [ ] https://github.com/edx/frontend-component-header-edx
    1. [ ] update `@edx/frontend-enterprise-catalog-search`, `@edx/frontend-enterprise-logistration`, `@edx/frontend-enterprise-utils` to use the same peer dependency
        - Example: https://github.com/openedx/frontend-enterprise/pull/278
    2. [ ] upgrade the release package to `frontend-component-header-edx`
        - Example: ....

**Post merge:**

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,21 @@ Describe what this pull request changes, and why. Include implications for peopl
 - [ ] Consider running your code modifications in the included example app within `frontend-platform`. This can be done by running `npm start` and opening http://localhost:8080.
 - [ ] Consider testing your code modifications in another local micro-frontend using local aliases configured via [the `module.config.js` file in `frontend-build`](https://github.com/openedx/frontend-build#local-module-configuration-for-webpack).
 - [ ] Verify your commit title/body conforms to the conventional commits format (e.g., `fix`, `feat`) and is appropriate for your code change. Consider whether your code is a breaking change, and modify your commit accordingly.
+  - [ ] Verify intended release version at https://github.com/semantic-release/semantic-release#how-does-it-work
+
+**For Major Release Only**
+- [ ] Update dependency/peer dependency repo to support the breaking change
+  - [ ] https://github.com/openedx/frontend-component-footer
+    - Example: https://github.com/openedx/frontend-component-footer/pull/241
+  - [ ] https://github.com/openedx/frontend-component-header
+    - Example: https://github.com/openedx/frontend-component-header/pull/275
+  - [ ] https://github.com/edx/frontend-component-footer-edx
+    - Example: https://github.com/edx/frontend-component-footer-edx/pull/208
+  - [ ] https://github.com/edx/frontend-component-header-edx
+    1. [ ] update `@edx/frontend-enterprise-catalog-search`, `@edx/frontend-enterprise-logistration`, `@edx/frontend-enterprise-utils` to use the same peer dependency
+        - Example: https://github.com/openedx/frontend-enterprise/pull/278
+    2. [ ] upgrade the release package to `frontend-component-header-edx`
+        - Example: ....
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,21 @@ Describe what this pull request changes, and why. Include implications for peopl
...
+    - Example: https://github.com/openedx/frontend-component-footer/pull/241
+  - [ ] https://github.com/openedx/frontend-component-header
+    - Example: https://github.com/openedx/frontend-component-header/pull/275
+  - [ ] https://github.com/edx/frontend-component-footer-edx
+    - Example: https://github.com/edx/frontend-component-footer-edx/pull/208
+  - [ ] https://github.com/edx/frontend-component-header-edx
+    1. [ ] update `@edx/frontend-enterprise-catalog-search`, `@edx/frontend-enterprise-logistration`, `@edx/frontend-enterprise-utils` to use the same peer dependency
+        - Example: https://github.com/openedx/frontend-enterprise/pull/278
+    2. [ ] upgrade the release package to `frontend-component-header-edx`
+        - Example: ....
```
arbrandes, line 24: Is there no PR to header-edx, yet?

arbrandes (changes requested):
@leangseu-edx, thanks for this!  I like the idea of listing downstream actions on breaking changes.

However, I have mixed feelings about including `edx` org-specifc checklist items on an `openedx` PR template: while we certainly don't want to break any downstream repos, to be fair we'd have to keep track of _every org's_ fork.  How about if instead a checklist item would say something like "Announce the breaking change in Slack and in the Forum"?

Would like to hear thoughts from @adamstankiewicz and @sarina, here.

.github/PULL_REQUEST_TEMPLATE.md
leangseu-edx: it require `frontend-enterprise` to update first before `header-edx` can update the `enterprise package and create the pr.

arbrandes: Alright, gotcha.  See my review comment, in any case: it probably makes this moot.

arbrandes, line 24: Had a conversation with the tCRIL team yesterday about this, and our position is this:

While tempting, having upstream (this repo) be responsible for downstream (frontend-component-*) is impractical: upstream can never know how many downstream repos there are, ultimately.

We feel the best solution here is for downstream to subscribe to upstream breaking changes instead.  This subscription could be theoretically automated via dependabot/renovate or some other bot, which would make the solution even better than a checklist.  What do you think?

To be clear, we think the **For Major Release Only** should be removed.  The "Verify intended release version" item is a good check to keep.

arbrandes (changes requested):
See inline comment.

.github/PULL_REQUEST_TEMPLATE.md
arbrandes: This is an example of the automation I mentioned above:

https://github.com/openedx/frontend-app-ecommerce/pull/217

leangseu-edx: That seems a reasonable approach.

PR #431, opened 2023-01-25
fix: used node 18 to fix release workflow which has latest semantic release

abdullahwaheed:
**Description:**

Used node 18 in `release` workflow which was breaking due to latest semantic release which has node 18 dependency.
Tested `npm ci`, `npm run build` and `npm run docs` locally 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,18 +14,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Node.js Env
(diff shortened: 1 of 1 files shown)


BilalQamar95 (approved):

arbrandes:
@abdullahwaheed, what is the impact of this on consumers of `frontend-platform` that do not yet support Node 18 explicitly?

abdullahwaheed:
@arbrandes we have just used node 18 in `release` workflow as newer version of `semantic release` dependent on node 18. It won't impact consumers as every consumer, which is a library, has its own release workflow

PR #440, opened 2023-02-15
feat: adds support for loading external theme CSS for MFEs

adamstankiewicz:
**Description:**

Implementation POC/prototype for ADR on loading a common, external stylesheet for MFEs.

* Introduces `useParagonTheme` in `AppProvider` to load/inject `<link>` elements for the core theme CSS and any theme variant CSS into the HTML document.
* Exposes the app theme state and a way for applications to expose functionality to toggle between theme variants to consumers via `AppContext`.
* Supports a dark theme variant (even though Paragon itself does not yet have a dark theme variant).
* Supports a few mechanisms for determining which theme variant to make "active" and visible to the user:
  * When an application sets a theme variant programmatically (e.g., by exposing a button to the user to switch between light and dark modes), the selected theme variant is stored in `localStorage` so it persists/loads across sessions and page refreshes.
  * When no theme variant is selected in `localStorage`, checks if the user's System Preferences has a preference for dark mode via a `@media` query for `prefers-color-scheme: dark`. If dark color scheme is preferred, the default configured dark theme variant is made visible, if one exists.
    * When the `prefers-color-scheme: dark` query changes (i.e., user updates their System Preferences), loads the appropriate default theme variant corresponding to the system preference in real time.
  * When either the dark mode theme variant doesn't exist or the default light theme is actively visible, finds the default configured light theme variant.

See [theming.md](https://github.com/openedx/frontend-platform/blob/e2a115edd2ce82d1855e63f0544d7c33f356c4f7/docs/how_tos/theming.md) for documentation and more details.

**Merge checklist:**

**Post merge:**


src/react/AppProvider.jsx
```diff
@@ -61,7 +68,28 @@ export default function AppProvider({ store, children }) {
...
+    locale,
+    appTheme: {
+      state: appThemeState,
+      dispatch: appThemeDispatch,
+    },
+  }), [authenticatedUser, config, locale, appThemeState, appThemeDispatch]);
+
+  if (!appThemeState?.isThemeLoaded) {
+    return null;
+  }
```
adamstankiewicz, line 92: [inform] Don't render the app until the core theme CSS and theme variant(s) CSS is loaded.

regisb:
I don't know enough about MFEs to comment about the details of this change, but what I understand is that it will require just two settings to customise the CSS, which is great. 

One thing, though: this new extension mechanism will only be as good as its documentation. So I urge you to spend some time to document it as best you can. Actually, I suggest that you write the docs _before_ finalizing the implementation, in [README-driven-development](https://tom.preston-werner.com/2010/08/23/readme-driven-development.html) style. For instance, you should describe about Tutor users can create and integrate new themes.

Reading the docs will allow us to comment on this PR in a much more informed way. Also, writing the docs will help you realize whether the new extension mechanism integrates well with the workflows of platform administrators and Open edX developers.

ghassanmas:
I have a couple of question, 
- If loading the theme/style from cdn, would that be enough for the app (to run UX/UI wise) without including the default app.*css file generated on each bundle. i.e. does the theme has enough css or just stuff to ride? 
- Will each MFE has to own file or one file is enough for all the MFEs for example app learning has its own scss file https://github.com/openedx/frontend-app-learning/blob/master/src/index.scss

arbrandes:
@adamstankiewicz, some stuff that came up out of a high-level discussion with tCRIL engineering:

* Support and longevity: we should consider the case where somebody is still running Palm after 5 years.  The required stylesheets must still be hosted somewhere, and the implication is also that they will have to be versioned.
* While we can envision the advantages of having themes be built separately and thus potentially reusable across versions, maybe this should be an optional behavior: say, if `APP_THEME_CORE_URL` is set, the stylesheet is fetched accordingly, but if not, then the app would expect to find it in the webpack bundle.  (This also addresses the issue of longevity.)
* Performance: there's nothing to say this proposal wouldn't be performant if done well, but we should keep UX in mind.  With the dynamic config API currently in place, it could mean making the user wait for two requests in sequence (not counting the very first GET), any of which could fail: first, for the configuration itself, and after this one is complete, for the stylesheets.  And this is not some theoretical possibility: it would be the default behavior for most Tutor deployments.

These aren't necessarily formal objections by tCRIL, by the way.   Just adding some stuff to keep in mind. :)

adamstankiewicz:
@arbrandes Thanks for sharing the input!

> Support and longevity: we should consider the case where somebody is still running Palm after 5 years.  The required stylesheets must still be hosted somewhere, and the implication is also that they will have to be versioned.

If we did end up using/recommending jsDelivr as an off-the-shelf free CDN supporting versions already aligned with NPM/Github releases, they guarantee any file from any version of the NPM package will remain accessible forever. They even support version fallback (say, a file is deleted in v1.0.2 that was previously there in v1.0.1, if you try to access that now-deleted file on v1.0.2, jsDelivr is smart enough to fallback to the version of the file in v1.0.1.

> While we can envision the advantages of having themes be built separately and thus potentially reusable across versions, maybe this should be an optional behavior: say, if APP_THEME_CORE_URL is set, the stylesheet is fetched accordingly, but if not, then the app would expect to find it in the webpack bundle. (This also addresses the issue of longevity.)

I was actually just chatting with @brian-smith-tcril about this a bit, too 😄 generally, though, +1 to the input and I'd extend it to say that we might be able to fallback to the installed CSS within the webpack bundle should the external CSS fail to load for whatever reason so at least the app renders with _some_ styles, even if those styles may be from a slightly older version than what might be upstream in the CDN, for example.

I believe this latter bit around falling back to locally installed CSS addresses at the third feedback item as well regarding performance? If the external CSS files are on a CDN(e.g., jsDelivr), the first few requests to a particular version of the theme may take a bit longer but subsequent requests to the same versions/files in the future would be cached and nearly instantaneous to load.

See https://www.jsdelivr.com/documentation for more details on how jsDelivr is production-ready (e.g., if Cloudflare goes down, jsDelivr instantly reverts to another CDN provider instead, has support for China, etc.).

adamstankiewicz:
@ghassanmas's questions above were largely answered in the Frontend Working Group meeting earlier today. To recap though:

> If loading the theme/style from cdn, would that be enough for the app (to run UX/UI wise) without including the default app.*css file generated on each bundle. i.e. does the theme has enough css or just stuff to override? I am referring https://github.com/openedx/frontend-build/pull/266#issuecomment-1428690893

The core theme CSS (`core.css`) intends to be an all-encompassing theme containing all of the Paragon styles necessary the foundational parts of the Open edX theme including layout, typography, component styles, etc. The light theme variant CSS (`light.css`) - in the future, possibly an additional `dark.css` for dark mode support - would define just the CSS variables necessary for the light theme variant (the default). These CSS variables are used within `core.css`.

Assuming applications don't have their own custom styles, applications indeed could not have any custom SCSS/CSS files at all, only relying on the external theme CSS.

If applications do have their own custom styles (hopefully not overriding Paragon styles!), `@edx/frontend-build` (via Webpack) will bundle the application's CSS to `.css` file(s) and inject it _after_ the external theme CSS.

By doing so, the application `.css` has higher priority over the theme CSS files and will (generally) take precedence.

> Will each MFE has to own file or one file is enough for all the MFEs for example app learning has its own scss file https://github.com/openedx/frontend-app-learning/blob/master/src/index.scss

The externally hosted, already-compiled theme CSS will need to versioning according to semantic versioning and aligned to NPM/Github releases. In our current MFE architecture, MFEs would use the latest release for the major version compatible with the installed `@edx/paragon` version in the MFE. For example, if the frontend-app-profile MFE has `@edx/paragon@21` installed in its package.json file, the MFE should also be using the latest `v21` of the external theme CSS to ensure compatibility.

adamstankiewicz:
> Reading the docs will allow us to comment on this PR in a much more informed way. Also, writing the docs will help you realize whether the new extension mechanism integrates well with the workflows of platform administrators and Open edX developers.

@regisb Noted! Yes, documentation is definitely a critical component of this project to move from current state of theming to runtime theming via design tokens. While there is no specific documentation to this frontend-platform implementation quite yet, the existing [ADR](https://github.com/openedx/frontend-platform/pull/351) proposing a related approach helped give some motivation/context for this PR. That said, I've taken an action item to:
* Write a brief ADR to go alongside this implementation in the same PR, likely borrowing certain aspects of the existing ADR linked above, describing the motivation/context, decision, etc.
* Include a "how to" doc in the repo with more specifics about "How do I use this theming extension?" for consumers.

felipemontoya:
I can't comment on the implementation, but the overall approach, and the concept of the core and light css variable bundles seem very appropriate to me. Thanks for getting this done.

regisb:
Thanks for your answer @adamstankiewicz! I think that the extension docs should live close to the rest of the developer docs: https://docs.openedx.org/en/latest/developers/index.html

xitij2000:
> The core theme CSS (`core.css`) intends to be an all-encompassing theme containing all of the Paragon styles necessary the foundational parts of the Open edX theme including layout, typography, component styles, etc. The light theme variant CSS (`light.css`) - in the future, possibly an additional `dark.css` for dark mode support - would define just the CSS variables necessary for the light theme variant (the default). These CSS variables are used within `core.css`.

Considering that it's now possible to have richer JSON-based config thanks to the config API. (i.e. the config API can return a JSON for any setting). Perhaps this code can be made more generic to load any kind of theme?

i.e. the config api can return a structure like:
```
{
  "core": "https...",
  "variants":  {
     "light": "https...",
     "dark": "https...",
     "high-contrast": "https..."
  }
}
```
Which is the structure being built here: 

https://github.com/openedx/frontend-platform/pull/440/files#diff-e986d374444ced0f5162cc6cb8b3dac6cbd228f339851f2a99a4aec545559d87R71-R78

If the config API isn't enabled, then it can continue to fall back to the `APP_THEME_LIGHT_URL`.

adamstankiewicz:
@xitij2000 Great suggestion! Yeah, agreed this should prefer any runtime config settings, and otherwise fall back to the regular ol' environment variables. I can't imagine it'd be too much of a lift to support the runtime config use case, to use an already created object structure representing the theme core/variant URLs, if it exists.

I'm not immediately sure what that implementation might look like as I'm not all that familiar with the config API outside of how frontend-platform calls it; is this something you might want to take a stab at (feel free to put some pseudocode or otherwise together as a PR to this one!)?

Out of curiosity... I'm not actually sure, is the runtime config used in production by anyone at this point (AFAIK it's not being used by 2U/edX yet)?

felipemontoya:
We are already using the runtime config in prod for one medium sized instance with ~70 multitenant sites. So far so good with its usage.

xitij2000:
> I'm not immediately sure what that implementation might look like as I'm not all that familiar with the config API outside of how frontend-platform calls it; is this something you might want to take a stab at (feel free to put some pseudocode or otherwise together as a PR to this one!)?

Sure! I'll put together something based on this PR and link it here when ready.

adamstankiewicz:
> Sure! I'll put together something based on this PR and link it here when ready.

@xitij2000 I ended up putting together a possible solution to prefer config from the MFE config API; otherwise, fallback to environment variables. I just pushed up the latest changes to this PR. The runtime config vs. environment variable config approaches are described in a new `docs/how_tos/theming.md` document.

arbrandes (commented):
This looks great, @adamstankiewicz!  Very exciting, actually!  Just made a couple of typo-level suggestions, and I did have a thought about how dynamic the dynamic config support will actually be.

src/react/hooks.js
```diff
@@ -21,3 +27,208 @@ export const useAppEvent = (type, callback) => {
...
+export const useParagonThemeCore = ({
+  coreThemeUrl,
+  onLoad,
+}) => {
+  useEffect(() => {
+    // If there is no config for the core theme url, do nothing.
+    if (!coreThemeUrl) {
+      return;
+    }
+    let coreThemeLink = document.head.querySelector(`link[href='${coreThemeUrl}']`);
```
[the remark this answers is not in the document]
xitij2000: I haven't tried it, but I think it should work? The code should update the stylesheet links such that the current theme is loaded, so it should respond to changes to runtime config and change of variant as long as there is a mechanism to call this code when such things happen.

docs/how_tos/theming.md
```diff
@@ -0,0 +1,40 @@
...
+
+`@edx/frontend-platform` additionally supports loading application configuration from an API at runtime rather than environment variables. For example, in `edx-platform`, there is an API endpoint for MFE runtime configuration at `http://localhost:18000/api/mfe_config/v1`. The application configuration may be setup via Django settings as follows:
+
+```
+ENABLE_MFE_CONFIG_API = True
+MFE_CONFIG = {}
+MFE_CONFIG_OVERRIDES = {
+    "profile": {
+        "PARAGON_THEME_URLS": {
+            'core': 'https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0-alpha.15/dist/paragon.css',
```
xitij2000, line 33: Would it make sense to support variable substitutions here? i.e. the URL could be:

``https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/paragon.css``

This way with each upgrade you don't need to update this value here in a completely different place.

dcoa:
I made some testing about this implementation and paragon alpha and I'm excited about it. I would like to ask about the next steps of this implementation, this draft pretends to become an open PR or do you have in mind a better solution that we can check?

Thanks for all the work around theming.

andrey-canon:
Hi @adamstankiewicz, I was testing this implementation and I found an error, when I set both variables `PARAGON_THEME_VARIANTS_LIGHT_URL`  and `PARAGON_THEME_CORE_URL`  the mfe doesn't loads 

Both variables

Screenshot from 2023-04-25 18-27-15


One variable

Screenshot from 2023-04-25 18-37-41

docs/how_tos/theming.md
dcoa: The last version of Paragon Alpha creates a folder instead of a file (core, and theme variant)

- @edx/paragon
    - styles
       - css
          - core
              - custom-media-breakpoints.css
              - index.css
              - variables.css
         - themes
           - light
               -  index.css
                - utility-classes.css
                - variables.css
               
I was wondering if this implementation should take that into consideration or should be resolved by the jsdelivr capacity to load more than one file per request https://www.jsdelivr.com/documentation#id-combine-multiple-files?

adamstankiewicz: > Would it make sense to support variable substitutions here? i.e. the URL could be: https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/paragon.css

@xitij2000 Can you expand a bit more on how you think the variable substitution would  work here? Where would `$paragonVersion` come from?

xitij2000: > > Would it make sense to support variable substitutions here? i.e. the URL could be: https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/paragon.css
> 
> @xitij2000 Can you expand a bit more on how you think the variable substitution would work here? Where would `$paragonVersion` come from?

The idea with variable substitution is that if you upgrade the version of paragon in the MFE it should automatically inject its version of paragon into this URL instead of requiring you to update the MFE configuration is site-configuration, or in settings. 

The version of Paragon to be filled in here is known at build time since it's in package.json so it can just be a static variable auto-discovered from the package.json file.

src/react/hooks.js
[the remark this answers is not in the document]
adamstankiewicz: @arbrandes @xitij2000 Correct, once the MFE runtime config cache expires after the default 5 minutes, it would start pulling in the updated stylesheet links. There is support for changing the theme variant (i.e., light vs. dark) at runtime by exposing a `paragonThemeActions` export from `@edx/frontend-platform/react`.

Example from a consumers POV:

```jsx
import React, { useContext } from 'react';
import { AppContext, paragonThemeActions } from '@edx/frontend-platform/react';

const LoadDarkMode = () => {
  const { paragonTheme } = useContext(AppContext);
  const { setParagonThemeVariant } = paragonThemeActions;
  return (
    <Button onClick={() => paragonTheme.dispatch(setParagonThemeVariant('dark'))}>
      Load dark mode
    </Button>
  );
};
```

adamstankiewicz:
@andrey-canon Appreciate you doing some additional QA on this branch! Do you happen to see any relevant JavaScript errors in the console? When I run this branch against `frontend-app-profile` using the CDN urls for the `@edx/paragon@alpha` release, it seems to be working as expected. See screenshot with both the `core.css` and `light.css` URLs injected into the `<head>`.

image

src/react/hooks.js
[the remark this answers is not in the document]
adamstankiewicz: @arbrandes I also just pushed up a commit to clean up the `<link>` elements for previously used versions should the MFE runtime config API return a different Paragon version specified in the CDN URL (e.g., should it ever get updated periodically without a full page refresh).

Example of switching versions at runtime, with no Flash of Unstyled Content (FoUC): 

https://github.com/openedx/frontend-platform/assets/2828721/d63674cb-fc6f-42eb-9243-69ecb22b957c

docs/how_tos/theming.md
adamstankiewicz: @xitij2000 That's a great suggestion, which should definitely mitigate the concern of needing to keep the CDN theme url versions in sync with the locally installed Paragon version.

To be able to support this, I believe we'd need to expose a `PARAGON_VERSION` environment variable from `@edx/frontend-build`, e.g.: https://github.com/openedx/frontend-build/pull/365

adamstankiewicz: [inform] I pushed up a commit using the above linked version of `@edx/frontend-build`.

ghassanmas: Assuming I am not incorrect, this would lead to style shift, right? because the flow as I understnad would as 
- The HTML of the is MFE is served 
- Browser start fetching resoucrs app.js .app.css ...etc
- Now concurrently browser would 
  - Execuse the app.js _of which it fetch the dynamic config api_ 
  -  Start rendering the screen
- Now if dynamic config from previous step would chane the style, there would be stlyle shift i.e. _user would see a theme style and then **suddenly** the theme/style would change_. 

Am I right, with the timeline of events above?, either cases I think the use PARAGON_VERSION is valuable, but we probably shall ask if sudden style change is okay and if not how to resolve it...

ghassanmas: The problem I am referring to is exactly this https://css-tricks.com/content-jumping-avoid/

ghassanmas: Reading into the thread, I think my comment above, is related even if we don't define `PARAGON_VERSION`. i.e even if just define `PARAGON_THEME_URLS`, _wouldn't the MFE at first load the deafult theme that was used/set when the MFE is built, before it get a chance to update it using via the dynamic config API_?.

adamstankiewicz: @ghassanmas I believe the intent here is that the MFE would no longer import any of Paragon's theme SCSS, in favor of having `@edx/frontend-platform` inject it via `AppProvider` at runtime, or otherwise fallback to the CSS included in the locally installed Paragon verison.

In this case, the MFE isn't shown until the externally loaded CSS files are injected/loaded by the browser, similar to how the MFE isn't rendered until the request to fetch the dynamic MFE config API is resolved.

See the below graphic for the flow in this proposed implementation for injecting/loading the Paragon theme URLs (I'll plan on adding a similar graphic to the ADR in this PR as well):

image

package.json
```diff
@@ -34,7 +34,7 @@
   "devDependencies": {
     "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
     "@edx/browserslist-config": "1.2.0",
-    "@edx/frontend-build": "12.8.27",
+    "@edx/frontend-build": "github:openedx/frontend-build#ags/2321",
```
adamstankiewicz, line 37: https://github.com/openedx/frontend-build/pull/365

package.json
```diff
@@ -74,8 +74,8 @@
     "universal-cookie": "4.0.4"
   },
   "peerDependencies": {
-    "@edx/paragon": ">= 10.0.0 < 21.0.0",
     "@edx/frontend-build": ">= 8.1.0",
```
adamstankiewicz, line 77: Will need to bump peer dependency of `@edx/frontend-build` once https://github.com/openedx/frontend-build/pull/365 merges.

src/react/hooks.js
```diff
@@ -48,3 +52,298 @@ export const useTrackColorSchemeChoice = () => {
...
+export const useParagonTheme = (config) => {
+  const paragonThemeUrls = getParagonThemeUrls(config);
+  const {
+    core: themeCore,
+    variants: themeVariants,
+  } = paragonThemeUrls;
+
+  const initialParagonThemeState = {
+    isThemeLoaded: false,
+    themeVariant: 'light',
```
adamstankiewicz, line 293: TODO: be smart about the default `themeVariant`. Use the `default` property from `paragonThemeUrls`.

src/react/hooks.js
```diff
@@ -48,3 +52,298 @@ export const useTrackColorSchemeChoice = () => {
...
+  };
+  const [themeState, dispatch] = useReducer(paragonThemeReducer, initialParagonThemeState);
+
+  const [isCoreThemeLoaded, setIsCoreThemeLoaded] = useState(false);
+  const [isLightThemeVariantLoaded, setIsLightThemeVariantLoaded] = useState(false);
```
adamstankiewicz, line 298: TODO: find a way to handle the loading state for any `variants` passed in, so we don't have to hardcode to specific theme variants.

package.json
davidjoy: This will presumably change prior to merging?

adamstankiewicz: @davidjoy Yes. the dependent `@edx/frontend-build` PR on associated with that `ags/2321` branch will need to be merged in prior to this PR merging.

docs/how_tos/theming.md
```diff
@@ -0,0 +1,76 @@
...
+            light: 'https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/light.css',
+        },
+    },
+};
+export default config;
+```
+
+### JavaScript-based configuration
+
+Another approach to configuration with `@edx/frontend-platform` is to create a `env.config.js` file in the root of the repository, similar to the environment variable configuration mentioned above. However, in this case, the configuration is defined as a JavaScript file, which affords consumers to use more complex data types than just strings as in the environment variable approach.
```
davidjoy, line 31: Note, I don't think the environment variable configuration mentioned above is actually mentioned above anymore. :) This paragraph may need to be revisited.

docs/how_tos/theming.md
```diff
@@ -0,0 +1,76 @@
...
+        },
+    },
+};
+export default config;
+```
+
+
+### MFE runtime configuration API
+
+`@edx/frontend-platform` additionally supports loading application configuration from an API at runtime rather than environment variables. For example, in `edx-platform`, there is an API endpoint for MFE runtime configuration at `http://localhost:18000/api/mfe_config/v1`. The application configuration may be setup via Django settings as follows:
```
davidjoy, line 50: Nit: To me this paragraph sort of implies that the MFE config API in edx-platform is one example of how this can be done, rather than the sole way frontend-platform can get this info.  (I mean, I guess someone could configure the frontend to point at some other service, but none exists!)

A link to the MFE config API doc here would probably be helpful.

src/react/AppProvider.jsx
```diff
@@ -63,7 +68,23 @@ export default function AppProvider({ store, children }) {
...
+    authenticatedUser,
+    config,
+    locale,
+    paragonTheme: {
+      state: paragonThemeState,
+      setThemeVariant: (themeVariant) => {
+        paragonThemeDispatch(paragonThemeActions.setParagonThemeVariant(themeVariant));
+      },
+    },
+  }), [authenticatedUser, config, locale, paragonThemeState, paragonThemeDispatch]);
```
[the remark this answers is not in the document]
davidjoy: Same sanity check question for `paragonThemeState`, I guess, though that'd be shocking if it weren't.

src/react/hooks.js
```diff
@@ -48,3 +52,298 @@ export const useTrackColorSchemeChoice = () => {
...
+
+      // create new link
+      const createCoreThemeLink = (url) => {
+        coreThemeLink = document.createElement('link');
+        coreThemeLink.href = url;
+        coreThemeLink.rel = 'stylesheet';
+        coreThemeLink.dataset.paragonThemeCore = true;
+        coreThemeLink.onload = () => {
+          onLoad();
+          removeExistingLinks(existingLinks);
```
davidjoy, line 92: What happens if the `onLoad` takes a while and we don't get to remove these existing links?  Does it adversely affect the layout at all?  (I'd assume not, but here I am just being paranoid)

src/react/hooks.js
```diff
@@ -48,3 +52,298 @@ export const useTrackColorSchemeChoice = () => {
...
+     * on if its the currently applied variant.
+     */
+    const generateStylesheetRelAttr = (themeVariant) => (currentThemeVariant === themeVariant ? 'stylesheet' : 'alternate stylesheet');
+
+    /**
+     * A helper function to determine which theme variant callback should be used
+     * based on the current theme variant.
+     */
+    const setThemeVariantLoaded = (themeVariant) => {
+      if (themeVariant === 'light') {
```
davidjoy, line 154: Is there a way we could use the set of theme variant keys in `PARAGON_THEME_URLS` in order to avoid hard-coding 'light', or needing to update frontend-platform when other variants are created?  If a variant doesn't exist, maybe there's just a fallback? 🤔  Seems like we should have enough info on what variants exist to avoid baking them in...

src/react/hooks.js
```diff
@@ -48,3 +52,298 @@ export const useTrackColorSchemeChoice = () => {
...
+        );
+      } else if (themeVariantLink.rel !== stylesheetRelForVariant) {
+        themeVariantLink.rel = stylesheetRelForVariant;
+      }
+    });
+  }, [themeVariants, currentThemeVariant, onLoadThemeVariantLight]);
+};
+
+const handleParagonVersionSubstitution = (url) => {
+  if (!url || !url.includes('$paragonVersion') || !PARAGON?.version) {
```
davidjoy, line 218: Alright, I'll bite.  I can't for the life of me figure out where `PARAGON` is coming from 😄

src/react/hooks.js
adamstankiewicz: @davidjoy Ended up creating a solution like this: https://github.com/openedx/frontend-platform/pull/440/files#diff-1947ea53b7d8a55f3e5e7c5a9bac408a3f20404f479e96782edfbde2ecc61d63R139-R146

```js
useEffect(() => {
  Object.keys(themeVariants || {}).forEach((themeVariant) => {
    setHasThemeVariantLoadedByName((prevState) => ({
      ...prevState,
      [themeVariant]: false,
    }));
  });
}, [themeVariants]);


useEffect(() => {
    if (Object.values(hasThemeVariantLoadedByName).some((hasLoaded) => hasLoaded)) {
      onLoad();
    }
}, [hasThemeVariantLoadedByName, onLoad]);
```

src/react/hooks.js
adamstankiewicz: See https://github.com/openedx/frontend-platform/pull/440#discussion_r1211073049

src/react/hooks.js
adamstankiewicz: Ha, magic no more: https://github.com/openedx/frontend-build/pull/365

docs/how_tos/theming.md
adamstankiewicz: > Nit: To me this paragraph sort of implies that the MFE config API in edx-platform is one example of how this can be done, rather than the sole way frontend-platform can get this info. (I mean, I guess someone could configure the frontend to point at some other service, but none exists!)

Agreed. Re-worded, and will push up a fix shortly.

> A link to the MFE config API doc here would probably be helpful.
@davidjoy Is there a suitable link to include here? I'm not aware of one. I know I started to create this ([currently 2U-internal](https://2u-internal.atlassian.net/wiki/spaces/FEDX/pages/387907737/Configuring+Micro-frontends+MFEs#Runtime-configuration-API)) document describing the different configuration mechanisms because I couldn't find any other documentation about the MFE runtime config API.

That being said, yes, a link to some more detailed, public docs here would be helpful :)

src/react/AppProvider.jsx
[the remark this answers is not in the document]
adamstankiewicz: Both `paragonThemeState` and `dispatch` from `useReducer` are safe to use in dependency arrays (memoized already like a state setter function).

From https://react.dev/reference/react/useReducer#usereducer:

> **If the new value you provide is identical to the current state, as determined by an [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will skip re-rendering the component and its children.** This is an optimization. React may still need to call your component before ignoring the result, but it shouldn’t affect your code.

From https://shopify.engineering/master-reacts-stable-values:

> useState and **useReducer return a state update function that is constant** – the hook will always return the same function. The value is guaranteed to be stable, because we know that the same object will be returned until we call the state update function.

docs/how_tos/theming.md
arbrandes: I think this is a good starting point:  https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/mfe_config_api/docs/decisions/0001-mfe-config-api.rst

src/react/hooks.js
adamstankiewicz: @davidjoy If `coreThemeLink.onload` is still loading, its styles haven't been applied in the browser yet. given this, I don't believe this is a concern? Any previous existing links would (and should) remain until the new link is fully resolved; otherwise, there would be a Flash of Unstyled Content.

For similar reasons, the `removeExistingLinks` is called after the `onLoad()` to ensure the new stylesheet is applied before removing the old ones

felipemontoya:
Coming back from the contributors meeting to bump this.

I have not been able to give a proper review because my frontend skills are not quite up to the task. However we have been using this idea to theme different instances in installations running nutmeg and olive. We backported the feature and the paragon design tokens to both releases. We are not looking into backporting to palm. I'd be very happy to make this land in master so that we can build more theming capabilities on top of it.

@dcoa please could you give this a new look with the last version changes and provide an official checkmark from our side if everything is working correctly?

adamstankiewicz:
@felipemontoya Appreciate the ping, and it was great catching up a bit with you earlier this week to understand the eduNext use cases in more depth!

As mentioned, the design tokens project is close to being ready-for-release, with the exception of some quality of life improvements targeted at improving the migration process for consumers across the Open edX platform (37+ repos), such as documentation and any opportunities for CLI tooling. I gave an estimate of perhaps conservatively 2-3 months 🤞 

As for this work in particular, I spent the afternoon updating it to account for the feedback from a recent Paragon Working Group where we discussed this PR to add support for a `defaults` configuration property in `PARAGON_THEME_URLS`.

The only thing I believe that's remaining for this PR at this point are tests 👀 

---

I've also added some additional features to this PR since we spoke:

* **Sets users' theme variant selection (e.g., light vs. dark) in `localStorage`** so the user's choice persists across sessions and page refreshes.

[paragon-dark-mode-localstorage-persist.webm](https://github.com/openedx/frontend-platform/assets/2828721/725b1208-95b5-43fc-b5cb-a934cf0ff408)

* If user has not made an explicit choice (i.e., no value stored in `localStorage`), **checks if the user's system preference prefers a dark color scheme, and if so, loads the dark theme variant (if configured).** Responds to changes in the system preferences in real time.

[paragon-dark-mode-system-preferences.webm](https://github.com/openedx/frontend-platform/assets/2828721/1dc7780d-3de0-45a6-afe8-96d69cb10e0a)

* Otherwise, **falls back to the configured default light theme variant.**

```json
{
    "core": {
        "url": "https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/core.min.css"
    },
    "defaults": {
        "light": "light",
    },
    "variants": {
        "light": {
            "url": "https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/light.min.css",
        }
    }
}
```

* Adds a `data-paragon-theme-variant="light|dark"` attribute to the `<html>` element, which gives consuming applications the ability to know the current theme variant via their own application CSS and JS (e.g., in a CSS file, where React Context is not available).

dcoa:
I was testing system preference configuration and it works when default and brandOverride are defined but when I use only default values like:

```
"PARAGON_THEME_URLS": {
                "core": {
                    "url": "https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0-alpha.38/dist/core.min.css"
                },
                "defaults": {
                    "dark": "dark",
                    "light": "light"
                },
                "variants": {
                    "dark": {
                        "url": "https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0-alpha.38/dist/dark.min.css""
                    },
                    "light": {
                        "url": "https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0-alpha.38/dist/light.min.css"
                    }
                }
            }
```
can not change the theme properly:
[system-preference-test.webm](https://github.com/openedx/frontend-platform/assets/66016493/3caa1e8c-b926-4644-ad63-9eede11de130)

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,196 @@
...
+              brandThemeVariantLink,
+            );
+          }
+        } else {
+          setIsBrandThemeVariantLoaded(true);
+        }
+      } else {
+        const updatedStylesheetRel = generateStylesheetRelAttr(themeVariant);
+        existingThemeVariantLink.rel = updatedStylesheetRel;
+        existingThemeVariantBrandLink.rel = updatedStylesheetRel;
```
dcoa, line 190: Need an extra conditional when brand link is not defined
```suggestion
        if (existingThemeVariantBrandLink) { existingThemeVariantBrandLink.rel = updatedStylesheetRel; }
```

src/react/hooks/paragon/useParagonThemeCore.js
```diff
@@ -0,0 +1,133 @@
...
+
+  useEffect(() => {
+    // If there is no config for the core theme url, do nothing.
+    if (!themeCore?.urls) {
+      setIsParagonThemeCoreLoaded(true);
+      setIsBrandThemeCoreLoaded(true);
+      return;
+    }
+    const getParagonThemeCoreLink = () => document.head.querySelector('link[data-paragon-theme-core="true"');
+    const existingCoreThemeLink = document.head.querySelector(`link[href='${themeCore.urls.default}']`);
```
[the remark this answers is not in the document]
xitij2000: I think a similar change will need to be applied to other queries as well.

Mashal-m:
Hey @adamstankiewicz , What is the current status of this PR, is it ready to review and merge? 
Could you please resolve conflicts?

src/react/hooks/useAppEvent.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param {string} type
+ * @param {function} callback
+ */
+const useAppEvent = (type, callback) => {
+  useEffect(() => {
+    const subscriptionToken = subscribe(type, callback);
+
+    return function cleanup() {
+      unsubscribe(subscriptionToken);
+    };
```
01001110J, line 22: Is possible to change this to an arrow function?

```suggestion
    return () => unsubscribe(subscriptionToken);
```

xitij2000:
@adamstankiewicz  Is there any way we can help get this closer to merging? We're currently using this in conjunction with patched versions of MFEs and it's working pretty well.

arbrandes:
This was discussed today in an FWG meeting,  It sounds like this will be taken forward as part of a Design Token Axim Funded Contribution.

adamstankiewicz:
Closed in favor of https://github.com/openedx/frontend-build/pull/546, which picked up where this PR left off :)

PR #458, opened 2023-03-20
fix: avoid multiple segment identify calls

shahbaz-arbisoft:
This PR adds a check to avoid multiple identify calls to the segment.

**Context:**

In Authn, a user has an anonymous user ID in all segment events until the user is authenticated. A user is moved to the welcome page after registration and we want to fire a segment identify call on the welcome page because the user is now authenticated. But the identify call is not fired until the user reloads the page manually. So we are explicitly calling `identifyAuthenticatedUser` in the welcome page so that the user has its actual user ID in all the segment events.

This works fine and identify call is fired with the user ID on the welcome page.

**Issue:**

The issue is when a user reloads the page, there are 3 identify calls being fired.

1. The automatically identify call due to reloading.
2. Our explicit identify call from `useEffect` which has `authenticatedUser` as a dependency which we get from the 
`getAuthenticatedUser` function.
3. We are also using `hydrateAuthenticatedUser` to attach some additional user attributes to authenticated User, which changes the `authenticatedUser` object causing `useEffect` to call `identifyAuthenticatedUser` again.

We want to avoid these multiple calls and we noticed that we have a `hasIdentifyBeenCalled` flag in `frontend-platform`, so we are using this flag in the `identifyAuthenticatedUser` function to avoid multiple identify calls.

### DIFF
--- a/src/analytics/SegmentAnalyticsService.js
+++ b/src/analytics/SegmentAnalyticsService.js
@@ -157,6 +157,10 @@ class SegmentAnalyticsService {
    * @param {*} [traits]
    */
   identifyAuthenticatedUser(userId, traits) {
+    if (this.hasIdentifyBeenCalled) {
+      return;
+    }
+
     if (!userId) {
       throw new Error('UserId is required for identifyAuthenticatedUser.');
     }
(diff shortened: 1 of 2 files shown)


davidjoy (commented):
@robrap Do we know of any reason why we might want to call identify more than once?  If not, this seems fine to me.  @shahbaz-arbisoft, I would perhaps suggest we do the same check in identifyAnonymousUser for parity.

shahbaz-arbisoft:
@davidjoy PR updated.

davidjoy (changes requested):
Oh, right, I forgot - we should add tests that prove we're not double-calling identify now.

shahbaz-arbisoft:
@davidjoy I have added tests for it, kindly review them.

PR #461, opened 2023-03-29
fix: styles not loading issue

Syed-Ali-Abbas-Zaidi:
**Description:**

This PR fixes the issue related to styles not being loaded.


abdullahwaheed:
please attach the ticket in the description and  fixed UIs screenshots as well

abdullahwaheed (approved):

example/index.jsx
```diff
@@ -12,10 +12,11 @@ import {
 import { APP_INIT_ERROR, APP_READY, initialize } from '@edx/frontend-platform';
 import { subscribe } from '@edx/frontend-platform/pubSub';
 
-import './index.scss';
 import ExamplePage from './ExamplePage';
 import AuthenticatedPage from './AuthenticatedPage';
 
+import('./index.scss');
```
adamstankiewicz, line 18: The dynamic import here makes the CSS get dynamically loaded at runtime, resulting in a Flash Of Unstyled Content ([FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content#:~:text=A%20flash%20of%20unstyled%20content,before%20all%20information%20is%20retrieved.)).

It also doesn't explain _why_ `import './index.scss';` doesn't work given that's how all MFEs and other example apps in our JS library repos import it without an issue, e.g. `frontend-component-footer` ([source](https://github.com/edx/frontend-component-header-edx/blob/21737a41b8617ba7e1f9fd233c58742c0b523eea/example/index.js#L10)).

adamstankiewicz (changes requested):

example/index.jsx
Syed-Ali-Abbas-Zaidi: I have looked into this issue and was not able to figure out why `import './index.scss';` is not working. Webpack config and version all the things are exactly similar to the ones used in `header/footer`.

However while doing RnD on this issue, I found [this](https://github.com/openedx/frontend-build/blob/master/config/webpack.dev.config.js#L67) comment in `webpack.dev.config` that `"We don't care about FOUC in development"`. As this dynamic import is being used in the example app, in my opinion we can ignore this issue for now. What do you suggest?

PR #462, opened 2023-03-30
revert: hasIdentifyBeenCalled check and its tests

shahbaz-arbisoft:
### Description:

This pull request reverses the earlier change where the flag `hasIdentifyBeenCalled` was added to the `identifyAuthenticatedUser` and `identifyAnonymousUser` functions, along with their respective tests. This modification proved to be ineffective in addressing the problem we faced and did not serve the intended purpose.

### Reason to revert:

The addition of a flag to the `identifyAuthenticatedUser` and `identifyAnonymousUser` functions caused a bug where the `identifyAnonymousUser` function was called with an anonymous ID on the registration page, setting the `hasIdentifyBeenCalled` flag to true. This caused the `identifyAuthenticatedUser` function was not called to perform the segment identify call after registration, as the `hasIdentifyBeenCalled` check was already true due to the previous function call. As a result, the segment identify call was not made and events were still being missed. This issue was discovered during testing, as it was not possible to test it locally, so the code had to be deployed on stage to identify the problem.

>  Reference PR #458

### DIFF
--- a/src/analytics/SegmentAnalyticsService.js
+++ b/src/analytics/SegmentAnalyticsService.js
@@ -157,10 +157,6 @@ class SegmentAnalyticsService {
    * @param {*} [traits]
    */
   identifyAuthenticatedUser(userId, traits) {
-    if (this.hasIdentifyBeenCalled) {
-      return;
-    }
-
     if (!userId) {
       throw new Error('UserId is required for identifyAuthenticatedUser.');
     }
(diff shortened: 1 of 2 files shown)


zainab-amir (approved):

shahbaz-arbisoft:
@davidjoy Would you mind reviewing my PR when you have a chance?

PR #466, opened 2023-04-04
fix: remove the hasIdentifyBeenCalled check and its corresponding tests

shahbaz-arbisoft:
### Description:

At first, a [PR](https://github.com/openedx/frontend-platform/pull/458) was created to address the issue of multiple segment identify calls. This was achieved by implementing a flag known as `hasIdentifyBeenCalled`, which checks whether the function has already been called and, therefore, eliminates the need for additional calls. 
Unfortunately, this solution did not fulfill our requirements, and we had to create a second [PR](https://github.com/openedx/frontend-platform/pull/462) to revert the changes. 
Although the revert PR was successfully merged into the master branch, the changes were not released due to the commit prefix. 

> All relevant information can be found in the linked PRs.

As a result, we created this new PR with an empty commit to ensure that the changes are released.


zainab-amir (approved):

zainab-amir (changes requested):
Please see discussion on slack

PR #471, opened 2023-04-05
[FC-0014] Add GA 4 to frontend-platform as separeted service component

UvgenGen:
**Description:**

Added support for GA4 to frontend-platform:
- Created new component GoogleAnalyticsService based on the existing SegmentAnalyticsService 

- Added tests for new component.

- The Google Analytics ID can be configured in the edx-platform MFE_CONFIG settings:
```
MFE_CONFIG = {
    "GOOGLE_ANALYTICS_4_ID": "G-TEST"
}
```
or, the ID can be set in the frontend-platform .env file as:
`GOOGLE_ANALYTICS_4_ID="G-TEST"`

**Merge checklist:**

**Post merge:**


openedx-webhooks:
Thanks for the pull request, @UvgenGen!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

davidjoy (changes requested):
First thing's first, this code is clean and correct.  Great job. :) 

The thing that makes me hesitate, though, is that we're creating this whole new interface/service just so that we can append a script to the head tag of the page.  It doesn't actually have any functionality that gets exposed to the app, or an actual 'interface' that the application uses that requires that we follow this interface/implementation pattern.  `analytics` and `logging` are generic wrappers around implementations (Segment and New Relic are the defaults), and they expose a bunch of functions that the app uses to perform various "analytics" and "logging" functions without knowing what the backing implementation is.

For those reasons, despite the code being very clear, I feel like we're choosing the wrong path... To add arbitrary scripts to the HTML, I'd expect to add code to frontend-build similar to [HtmlWebpackNewRelicPlugin](https://github.com/openedx/frontend-build/blob/master/lib/plugins/html-webpack-new-relic-plugin/HtmlWebpackNewRelicPlugin.js#L8).  

That HtmlWebpackNewRelicPlugin code appends an inline script tag, but an even simpler version could easily add GA to the header without the application runtime (i.e., frontend-platform) needing to be involved at all.  We could also generalize it for any arbitrary script someone wants to add - it doesn't necessarily need to be GA-specific long term.

src/google-analytics/GoogleAnalyticsService.js
```diff
@@ -0,0 +1,53 @@
...
+      const first = document.getElementsByTagName('script')[0];
+      first.parentNode.insertBefore(scriptSrc, first);
+      first.parentNode.insertBefore(scriptGtag, first);
+      googleAnalytics._loadOptions = options; // eslint-disable-line no-underscore-dangle
+    };
+
+    // Load GoogleAnalytics with your key.
+    googleAnalytics.load(this.analyticsId);
+  }
+}
```
arbrandes, line 51: Are we not going to implement artificial events such as [sendPageEvent()](https://github.com/openedx/frontend-platform/blob/master/src/analytics/SegmentAnalyticsService.js#L234-L241)?

I don't want to unwittingly extend scope, here.  It's just that I originally imagined a drop-in solution for replacing Segment with GA4.  But that's of course more complicated than just adding GA4 tags into MFEs that invoke this, which currently none do.

In other words, to use GA4 in MFEs with this new service, they still have to be forked or changed to do so.

arbrandes (commented):
Curious about how the feature can be used by operators in practice.

arbrandes:
I'm with @davidjoy, except for the fact that I'd much rather we implement the option of choosing between Segment and GA (and any future services) while getting the same features (i.e. artificial events) without having to modify MFEs.

Of course, _ideally_ we'd also have some way to make this pluggable, so people can implement an interface for their own service and have it be enabled without the need to modify `frontend-platform`, but this is definitely outside the scope, here. :)

src/google-analytics/GoogleAnalyticsService.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * @implements {GoogleAnalyticsService}
+ * @memberof module:GoogleAnalytics
+ */
+class GoogleAnalyticsService {
+  constructor({ config }) {
+    this.analyticsId = config.GOOGLE_ANALYTICS_4_ID;
```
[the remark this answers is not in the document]
bmtcril: That's actually the situation we're in right now. GA universal analytics is the current version and we can't just overwrite it because both the javascript snippet and identifier change with the upgrade. Because a handful of installs can continue using universal analytics longer than most of the community due to being 360 subscribers we need to be able to support both for a period of time. Once GA UA 360 reaches end of life we can deprecate and remove the old version.

src/google-analytics/GoogleAnalyticsService.js
bmtcril: Yeah for this work we're focusing on 1:1 functionality with GA universal, not Segment. I believe 2U has work coming down later this year to extend the functionality closer to how Segment is working, but I don't know how far investigation has gone in that direction.

arbrandes:
@davidjoy, just talked with @bmtcril about this, and while the build-time config makes sense. there's a catch: it won't allow operators to change the GA id based on `site_configuration`.  So it has to be dynamic.

As I mentioned earlier, I'd rather have a Segment drop-in replacement that uses the same service interface, but we can do that later - probably as part of making this truly pluggable.  I figure that for now, at least, this implementation should give many operators what they need out of the box.

cmltaWt0:
> 

@davidjoy  thanks for the valuable suggestion.

I sincerely agree with the simplicity approach. We evaluated different approaches intending to simplify the solution and simultaneously being able to extend the features in the future.

The frontend build repo was one of the options we considered, but we decided to reject it due to the dynamic/runtime configuration requirement.
Another discussed option was the Helmet component usage, but if we had chosen it, we would have introduced a lot of complexity.

The new Service interface doesn't seem to be needed for a simple script injection, but the idea was to let the community and operators easily extend the Analytics Service (e.g. to add the GA Custom Event build methods with default Data Layer implementations).
Regarding the current state of introducing the new interface. Currently, the Segment interface implements TrackingLogs events, and we must leave the ability to have two analytics in the active state.

In the future, we can refactor the Analytics service to introduce the real pluggability.

davidjoy:
I think we need to agree on our requirements a bit.  

I understand why it can't be in frontend-build now, given that we want to configure the ID at runtime and not bake it into the app.  That's something we could probably work around, but then the code would be split between two repos which is probably worse.

So if we intend to implement the `analytics` interface with Google Analytics in the future, then adding this as an analytics service implementation in this repo is the right way to go.  But that would mean it should be an alternate implementation for the `analytics` service, not a new service type named `google-analytics`.  

If we _don't_ intend to implement the `analytics` interface functions, then this can't be an implementation for that interface because MFEs will blow up when there's no implementation of functions like `sendPageEvent`.

I'm also hearing that someone might want _both_ Segment and GA active at the same time?  My understanding was that Segment starts up GA on its own?  I'm not sure I understand why we'd want both simultaneously, though I'm happy to be wrong.

If that is a thing, then we need to just think through what "type of thing" GA is here w/r/t MFE initialization.  Is it just a script we want to load?  If so, then a generic 'script loader' step of some sort is probably the way to go.  If it fulfills some other function for MFEs and we want them to call it, then it _might_ be a new 'interface' type.  If it is, we wouldn't want to call it "google analytics" regardless - that's the name of an implementation, not of a generic interface.

I don't want to stand on ceremony too much where we're making reversible decisions, but some intentional architecture here is probably important.  Some of these choices aren't super easy to change later.

cmltaWt0:
@davidjoy 
We discussed with @GlugovGrGlib the possibility of re-implementing the current approach and the reasons why it is necessary. Our decision is that we can't advocate for a new interface because we don't know what features should be included in the Google Analytics Service in the future. I agree with you that adopting the new service could be an option, but it is also sufficient to implement a script loader without introducing a new interface and improve it in the future. Later, it can be even upgraded to a new pluggable solution.

@UvgenGen  will develop an alternative, simplified implementation (in a separate DRAFT PR) for us to review and analyze.

Regarding the Segment and GA operating simultaneously, I previously believed that this was the case for some customers, but now I am uncertain as well.

davidjoy:
I reviewed #472 - looks awesome.  Great work.

openedx-webhooks:
@UvgenGen Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-platform&entry.867055334=https://github.com/openedx/frontend-platform/pull/471&entry.1484655318=https://github.com/UvgenGen&entry.752974735=2023-04-05+14:13&entry.1917517419=2023-04-12+10:21&entry.2133058324=No) so we can improve your experience in the future.

PR #474, opened 2023-04-11
feat: JavaScript file configuration of apps

davidjoy:
This adds the ability to configure an application via an env.config.js file rather than environment variables or .env files.  This mechanism is much more flexible and powerful than environment variables as described in the associated ADR.

It also improves documentation around how to configure applications, providing more detail on the various methods.

Finally, it allows the logging, analytics, and auth services to be configured via the configuration document now that we can supply an alternate implementation in an env.config.js file.  This allows configuration of these services without forking the MFE.

It doesn't remove the ability to supply a service implementation via the `initialize()` function's arguments since that would technically be a breaking change and it could conceivably be in use in forks today.  At a later date we could choose to remove that mechanism.

fixes https://github.com/openedx/wg-frontend/issues/11

**Description:**

Describe what this pull request changes, and why. Include implications for people using this change.

**Merge checklist:**

- [x] Consider testing your code modifications in another local micro-frontend using local aliases configured via [the `module.config.js` file in `frontend-build`](https://github.com/openedx/frontend-build#local-module-configuration-for-webpack).

**Post merge:**

### DIFF
--- a/docs/decisions/0007-javascript-file-configuration.rst
+++ b/docs/decisions/0007-javascript-file-configuration.rst
@@ -28,7 +28,7 @@ token that helps the build replace it with a string literal.
 This approach has several important limitations:
 
 - There's no way to add variables without hard-coding process.env.XXXX somewhere in the file,
-  complicating our ability to add additional application-specific configuration without explici tly
+  complicating our ability to add additional application-specific configuration without explicitly
   merging it into the configuration document after it's been created in frontend-platform.
 - The method can *only* handle strings.
 
--- a/docs/decisions/0007-javascript-file-configuration.rst
+++ b/docs/decisions/0007-javascript-file-configuration.rst
@@ -1,0 +1,111 @@
+Promote JavaScript file configuration and deprecate environment variable configuration
+======================================================================================
+
+Status
+------
+
+Draft
+
+Context
+-------
+
+Our webpack build process allows us to set environment variables on the command line or via .env
+files.  These environment variables are available in the application via ``process.env``.
+
+The implementation of this uses templatization and string interpolation to replace any instance of
+``process.env.XXXX`` with the value of the environment variable named ``XXXX``.  As an example, in our
+source code we may write::
+
+    const LMS_BASE_URL = process.env.LMS_BASE_URL;
+
+After the build process runs, the compiled source code will instead read::
+
+    const LMS_BASE_URL = 'http://localhost:18000';
+
+Put another way, `process.env` is not actually an object available at runtime, it's a templatization
+token that helps the build replace it with a string literal.
+
+This approach has several important limitations:
+
+- There's no way to add variables without hard-coding process.env.XXXX somewhere in the file,
+  complicating our ability to add additional application-specific configuration without explici tly
+  merging it into the configuration document after it's been created in frontend-platform.
+- The method can *only* handle strings.
+
+  Other data types are converted to strings::
+
+    # Build command:
+    BOOLEAN_VAR=false NULL_VAR=null NUMBER_VAR=123 npm run build
+
+    ...
+
+    // Source code:
+    const BOOLEAN_VAR = process.env.BOOLEAN_VAR;
+    const NULL_VAR = process.env.NULL_VAR;
+    const NUMBER_VAR = process.env.NUMBER_VAR;
+
+    ...
+
+    // Compiled source after the build runs:
+    const BOOLEAN_VAR = "false";
+    const NULL_VAR = "null";
+    const NUMBER_VAR = "123";
+
+  This is not good!
+
+- It makes it very difficult to supply array and object configuration variables, and unreasonable to
+  supply class or function config since we'd have to ``eval()`` them.
+
+Related to all this, frontend-platform has long hand the ability to replace the implementations of
+its analytics, auth, and logging services, but no way to actually *configure* the app with a new
+implementation.  Because of the above limitations, there's no reasonable way to configure a
+JavaScript class via environment variables.
+
+Decision
+--------
+
+For the above reasons, we will deprecate environment variable configuration in favor of JavaScript
+file configuration.
+
+This method makes use of an ``env.config.js`` file to supply configuration variables to an application::
+
+    const config = {
+      LMS_BASE_URL: 'http://localhost:18000',
+      BOOLEAN_VAR: false,
+      NULL_VAR: null,
+      NUMBER_VAR: 123
+    };
+
+    export default config;
+
+This file is imported by the frontend-build webpack build process if it exists, and expected by
+frontend-platform as part of its initialization process. If the file doesn't exist, frontend-build
+falls back to importing an empty object for backwards compatibility.  This functionality already exists
+today in frontend-build in preparation for using it here in frontend-platform.
+
+This interdependency creates a peerDependency for frontend-platform on `frontend-build v8.1.0 <frontend_build_810_>`_ or
+later.
+
+Using a JavaScript file for configuration is standard practice in the JavaScript/node community.
+Babel, webpack, eslint, Jest, etc., all accept configuration via JavaScript files (which we take
+advantage of in frontend-build), so there is ample precedent for using a .js file for configuration.
+
+In order to achieve deprecation of environment variable configuration, we will follow the
+deprecation process described in `OEP-21: Deprecation and Removal <oep21_>`_.  In addition, we will
+add build-time warnings to frontend-build indicating the deprecation of environment
+variable configuration.  Practically speaking, this will mean adjusting build processes throughout
+the community and in common tools like Tutor.
+
+Rejected Alternatives
+---------------------
+
+Another option was to use JSON files for this purpose.  This solves some of our issues (limited use
+of non-string primitive data types) but is otherwise not nearly as expressive for flexible as using
+a JavaScript file directly.  Anecdotally, in the past frontend-build used JSON versions of many of
+its configuration files (Babel, eslint, jest) but over time they were all converted to JavaScript
+files so we could express more complicated configuration needs.  Since one of the primary use cases
+and reasons we need a new configuration method is to allow developers to supply alternate implementations
+of frontend-platform's core services (analytics, logging), JSON was a effectively a non-starter.
+
+.. _oep21: https://docs.openedx.org/projects/openedx-proposals/en/latest/processes/oep-0021-proc-deprecation.html
+.. _frontend_build_810: https://github.com/openedx/frontend-build/releases/tag/v8.1.0
--- a/docs/decisions/0007-javascript-file-configuration.rst
+++ b/docs/decisions/0007-javascript-file-configuration.rst
@@ -4,7 +4,7 @@ Promote JavaScript file configuration and deprecate environment variable configu
 Status
 ------
 
-Draft
+Accepted
 
 Context
 -------
--- a/docs/decisions/0007-javascript-file-configuration.rst
+++ b/docs/decisions/0007-javascript-file-configuration.rst
@@ -9,12 +9,13 @@ Draft
 Context
 -------
 
-Our webpack build process allows us to set environment variables on the command line or via .env
-files.  These environment variables are available in the application via ``process.env``.
+Our webpack build process allows us to set environment variables on the command
+line or via .env files.  These environment variables are available in the
+application via ``process.env``.
 
-The implementation of this uses templatization and string interpolation to replace any instance of
-``process.env.XXXX`` with the value of the environment variable named ``XXXX``.  As an example, in our
-source code we may write::
+The implementation of this uses templatization and string interpolation to
+replace any instance of ``process.env.XXXX`` with the value of the environment
+variable named ``XXXX``.  As an example, in our source code we may write::
 
     const LMS_BASE_URL = process.env.LMS_BASE_URL;
 
@@ -22,14 +23,16 @@ After the build process runs, the compiled source code will instead read::
 
     const LMS_BASE_URL = 'http://localhost:18000';
 
-Put another way, `process.env` is not actually an object available at runtime, it's a templatization
-token that helps the build replace it with a string literal.
+Put another way, `process.env` is not actually an object available at runtime,
+it's a templatization token that helps the build replace it with a string
+literal.
 
 This approach has several important limitations:
 
-- There's no way to add variables without hard-coding process.env.XXXX somewhere in the file,
-  complicating our ability to add additional application-specific configuration without explicitly
-  merging it into the configuration document after it's been created in frontend-platform.
+- There's no way to add variables without hard-coding process.env.XXXX
+  somewhere in the file, complicating our ability to add additional
+  application-specific configuration without explicitly merging it into the
+  configuration document after it's been created in frontend-platform.
 - The method can *only* handle strings.
 
   Other data types are converted to strings::
@@ -53,21 +56,24 @@ This approach has several important limitations:
 
   This is not good!
 
-- It makes it very difficult to supply array and object configuration variables, and unreasonable to
-  supply class or function config since we'd have to ``eval()`` them.
+- It makes it very difficult to supply array and object configuration
+  variables, and unreasonable to supply class or function config since we'd
+  have to ``eval()`` them.
 
-Related to all this, frontend-platform has long hand the ability to replace the implementations of
-its analytics, auth, and logging services, but no way to actually *configure* the app with a new
-implementation.  Because of the above limitations, there's no reasonable way to configure a
-JavaScript class via environment variables.
+Related to all this, frontend-platform has long had the ability to replace the
+implementations of its analytics, auth, and logging services, but no way to
+actually *configure* the app with a new implementation.  Because of the above
+limitations, there's no reasonable way to configure a JavaScript class via
+environment variables.
 
 Decision
 --------
 
-For the above reasons, we will deprecate environment variable configuration in favor of JavaScript
-file configuration.
+For the above reasons, we will deprecate environment variable configuration in
+favor of JavaScript file configuration.
 
-This method makes use of an ``env.config.js`` file to supply configuration variables to an application::
+This method makes use of an ``env.config.js`` file to supply configuration
+variables to an application::
 
     const config = {
       LMS_BASE_URL: 'http://localhost:18000',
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
   },
   "peerDependencies": {
     "@edx/paragon": ">= 10.0.0 < 21.0.0",
+    "@edx/frontend-build": ">= 8.1.0",
```
davidjoy, line 78: This is the version of frontend-build that enables `env.config` as a magic import.  So we didn't have a peer dependency here before, but we technically do now.  It's unlikely to bite anyone since 8.1.0 was quite a while ago

src/initialize.js
```diff
@@ -130,13 +139,32 @@ export async function auth(requireUser, hydrateUser) {
...
+ * This env.config.js is loaded at runtime and must export one of two things:
+ *
+ * - An object which will be merged into the application config via `mergeConfig`.
+ * - A function which returns an object which will be merged into the application config via
+ * `mergeConfig`.  This function can return a promise.
+ */
+async function jsFileConfig() {
+  let config = {};
+  if (typeof envConfig === 'function') {
+    config = await envConfig();
```
davidjoy, line 154: Hm, getting coverage on all these lines is challenging.  I'm not sure how I'd change the contents of the env.config file on a per-test basis. 🤔

src/initialize.js
```diff
@@ -130,13 +139,32 @@ export async function auth(requireUser, hydrateUser) {
...
+}
+
 /*
  * Set or overrides configuration through an API.
  * This method allows runtime configuration.
  * Set a basic configuration when an error happen and allow initError and display the ErrorPage.
  */
-
-export async function runtimeConfig() {
+async function runtimeConfig() {
```
davidjoy, line 167: I took away the `export` on this line intentionally, as this is an internal function and nothing is consuming the export.

docs/decisions/0007-javascript-file-configuration.rst
```diff
@@ -0,0 +1,111 @@
...
+
+    const LMS_BASE_URL = 'http://localhost:18000';
+
+Put another way, `process.env` is not actually an object available at runtime, it's a templatization
+token that helps the build replace it with a string literal.
+
+This approach has several important limitations:
+
+- There's no way to add variables without hard-coding process.env.XXXX somewhere in the file,
+  complicating our ability to add additional application-specific configuration without explici tly
```
dcoa, line 31:
```suggestion
  complicating our ability to add additional application-specific configuration without explicitly
```

src/initialize.js
adamstankiewicz: Same here, unfortunately. Just tried to play around with Jest mocks after pulling down the branch. I'm also not having much luck trying to mock the `env.config` import on a per-test basis... I was able to mock the import to always return a function, but then we don't get coverage on the non-function case. I was also able to mock it to return a custom object but then I couldn't find a way to change that mock object to a function in a second test.

kdmccormick:
I do not have my head fully wrapped around this, but I trust the reviews of those that do.

With my Build-Test-Release hat on, though, the questions that the ADR leaves me with are: Does this jive well with runtime configuration, which is now required by the community release? And are you happy with how runtime config is implemented, now that there's a deprecation path for `.env`-based configuration?

docs/decisions/0007-javascript-file-configuration.rst
```diff
@@ -0,0 +1,111 @@
...
+Related to all this, frontend-platform has long hand the ability to replace the implementations of
+its analytics, auth, and logging services, but no way to actually *configure* the app with a new
+implementation.  Because of the above limitations, there's no reasonable way to configure a
+JavaScript class via environment variables.
+
+Decision
+--------
+
+For the above reasons, we will deprecate environment variable configuration in favor of JavaScript
+file configuration.
```
adamstankiewicz, line 68: We might want to keep the legacy environment variable configuration around and not deprecate it. There's not much reason to deprecate it that I can think of (e.g., not much maintenance overhead to support it), though it'll probably start to get used less with the introduction of `env.config` and the MFE runtime config API.

We could observe usage over time and, if we see that very few MFEs use the environment variable configuration mechanism then, deprecate at that time.

davidjoy: I hear you - the thing that comes to mind is that if we start using env.config.js so that we can pass real data types into the app instead of just strings, we'll presumably start removing the code that converts `"false"` to `false`, etc., in which case for given MFEs the old config won't work properly anymore.  Unless we just make the logic _more_ complicated to accommodate both. 😬  I guess this is a problem even in the interim before an official deprecation/removal.  Hm - will have to think about this more.

src/initialize.js
```diff
@@ -253,18 +281,28 @@ export async function initialize({
...
     publish(APP_CONFIG_INITIALIZED);
 
+    // This allows us to replace the implementations of the logging, analytics, and auth services
+    // based on keys in the ConfigDocument.  The JavaScript File Configuration method is the only
+    // one capable of supplying an alternate implementation since it can import other modules.
+    // If a service wasn't supplied we fall back to the default parameters on the initialize
+    // function signature.
+    const loggingServiceImpl = getConfig().loggingService || loggingService;
+    const analyticsServiceImpl = getConfig().analyticsService || analyticsService;
+    const authServiceImpl = getConfig().authService || authService;
```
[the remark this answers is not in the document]
davidjoy: Oh, good thought!  Yes, that's a good idea.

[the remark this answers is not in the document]
davidjoy: Oh - I forgot, frontend-platform already does this. 😁  When a service is configured it compares it to an expected service shape using PropTypes.

src/config.js
```diff
@@ -2,23 +2,89 @@
...
  *
  * The configuration module provides utilities for working with an application's configuration
- * document (ConfigDocument).  This module uses `process.env` to import configuration variables
- * from the command-line build process.  It can be dynamically extended at run-time using a
- * `config` initialization handler.  Please see the Initialization documentation for more
- * information on handlers and initialization phases.
+ * document (ConfigDocument).  Configuration environment variables can be supplied to the
+ * application in four different ways:
+ *
+ * ##### Build-time Environment Variables
```
brian-smith-tcril, line 8: I'm not sure the best way to communicate this in this file, but I think it'd be nice to categorize the configuration methods

* Build-time configuration
  * Environment Variables (intend to deprecate)
  * JavaScript File
    * Object
    * Function returning object
    * Function returning promise
* Runtime configuration

src/config.js
```diff
@@ -2,23 +2,89 @@
...
+ *     resolve({
+ *       LMS_BASE_URL: 'http://localhost:18000'
+ *     });
+ *   });
+ * }
+ *
+ * export default getAsyncConfig;
+ * ```
+ *
+ * ##### Runtime Configuration
```
brian-smith-tcril, line 60: could be nice to add something explaining how this interacts with https://github.com/openedx/frontend-platform/pull/474/files#diff-23821ae067a9c7dfb6227774aefb39fa3951a5874ea953b9401dfc80175c026cR257-R259
> When using Webpack (i.e., normal usage), the build process is responsible for supplying these
> variables via command-line environment variables.  That means they must be supplied at build
> time.

brian-smith-tcril (commented):
This looks really cool!

src/config.js
kdmccormick: +1

src/config.js
davidjoy: Yes!  These are good thoughts.  There's an order that they get applied that I don't think we've called out anywhere.

src/config.js
davidjoy: I'm down with that, yeah!

src/config.js
```diff
@@ -2,23 +2,89 @@
  * #### Import members from **@edx/frontend-platform**
  *
  * The configuration module provides utilities for working with an application's configuration
- * document (ConfigDocument).  This module uses `process.env` to import configuration variables
- * from the command-line build process.  It can be dynamically extended at run-time using a
- * `config` initialization handler.  Please see the Initialization documentation for more
- * information on handlers and initialization phases.
+ * document (ConfigDocument).  Configuration environment variables can be supplied to the
+ * application in four different ways:
```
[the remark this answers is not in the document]
davidjoy: So many good suggestions. (I read these from the bottom up, so this is the third great suggestion I've seen. 😉)

package.json
davidjoy: For the record (Adam and I talked about this a bit offline) - the only way I could get `env.config` to work as an import was to use specialized webpack config (in frontend-build) to fallback to an empty import if the file couldn't be found.  Since the entire mechanism is predicated on that behavior, and it's implemented in frontend-build, it feels a bit unavoidable to add the peer dependency.  There's just not a way I've ever found to do this without that added magic (which was merged into frontend-build some time ago in https://github.com/openedx/frontend-build/pull/200)

docs/decisions/0007-javascript-file-configuration.rst
davidjoy: On second thought, it may not be too much of a problem.  The MFEs tend to have arguments in their calls to `initialize` where they pull config out of `process.env` if it exists.  The new `jsFileConfig` function runs _after_ that.  The first can continue to massage strings into other data types, and the second doesn't have to worry about it.  They can co-exist.

src/config.js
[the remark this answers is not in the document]
davidjoy: I've added some more deets.  Going to close this as resolved.

src/config.js
davidjoy: I reorganized it a bit to match.  Resolving!

src/config.js
davidjoy: I added some words around this and clarified why you'd use one or the other.  Resolving!

davidjoy:
@kdmccormick and @adamstankiewicz - so I think after reading through everything here, I'm _not_ going to back off the deprecation of environment variable config.  I'm also not 100% sure I have the will to fully follow through on getting rid of it myself. I think this probably looks like filing a DEPR ticket with all the deets on how we get rid of it, yeah?  Just trying to think through how to deliver on the intention in the ADR. 🤔

arbrandes (approved):
This is great!  Thanks David!

kdmccormick:
@davidjoy 
> I'm not going to back off the deprecation of environment variable config. I'm also not 100% sure I have the will to fully follow through on getting rid of it myself. I think this probably looks like filing a DEPR ticket with all the deets on how we get rid of it, yeah? Just trying to think through how to deliver on the intention in the ADR.

Sounds good! If you are able to write a DEPR ticket with removal steps, and then nudge it in the direction of actually happening (eg, emitting a warning when static config is used), then I think it's pretty likely someone in Frontend WG would eventually be able to drive the removal.

src/initialize.js
davidjoy: I figured it out.  I put each test in its own file and `jest.mock` for `env.config.js` with different implementations worked like a charm. :)  It also let me remove the real env.config.js file from the repo!  

I found some instructions online for doing it all in one file, but it didn't work for me, so I figure this will be decent for now.

davidjoy:
So I'm going to merge this first, _then_ create the DEPR ticket if only so I can link to the ADR in this PR once it hits its final home on master.

adamstankiewicz:
My only concern with the deprecation of the `.env`-based config approach is that it's the mechanism we use to support changing the default config in `.env.development` during local development within different settings that we don't want to end up committed, e.g. Algolia API keys/secrets.

In a world with only MFE runtime config and `env.config.js`, I believe we'd have to solely rely on the MFE runtime config API to create "private" config settings that we don't want to put into `env.config.js` to avoid the risk of getting it committed. 

This becomes a developer experience issue in that I need to have the MFE runtime config API configured in edx-platform to be able to do any development with my private config setting in the MFE, whereas before we could just create a `.env.private` file to avoid needing to be coupled to the MFE runtime config API during development.

Is there a possibility we could support of `env.config.private.js` or similar to avoid this coupling during local development? This just came up in at least 2-3 conversations in the last 2 weeks where it was much easier to help someone create a private environment variable via `.env.private` versus needing to do a big knowledge share around setting up the MFE runtime config API locally.

package.json
adamstankiewicz: FWIW, we may have ended up needing to do something similar given the direction [this PR](https://github.com/openedx/frontend-platform/pull/440) is headed to be able to fallback to a local Paragon version in the externally hosted Paragon CSS urls or if a externally hosted Paragon CSS urls fails to load, we need to expose the paths to the locally installed Paragon theme via frontend-build such that frontend-platform has access to them (otherwise, each MFE would have to pass args to `initialize` which feels like could be avoided).

src/config.js
```diff
@@ -2,23 +2,122 @@
...
+ * variables or JS file).  Configuration defined in a JS file will override environment variables.
+ *
+ * ##### Build-time Configuration
+ *
+ * Build-time configuration methods add config variables into the app when it is built by webpack.
+ * This saves the app an API call and means it has all the information it needs to initialize right
+ * away.  There are two methods of supplying build-time configuration: environment variables and a
+ * JavaScript file.
+ *
+ * ######Environment Variables
```
adamstankiewicz, line 25: nit: missing space between "######" and "Environment Variables"

src/initialize.js
```diff
@@ -46,6 +46,15 @@
...
 
 import { createBrowserHistory, createMemoryHistory } from 'history';
+/*
+This 'env.config' package is a special 'magic' alias in our webpack configuration in frontend-build.
+It points at an `env.config.js` file in the root of an MFE's repository if it exists and falls back
+to an empty object `{}` if the file doesn't exist.  This acts like an 'optional' import, in a sense.
+Note that the env.config.js file in frontend-platform's root directory is NOT used by the actual
+initialization code, it's just there for the test suite and example application.
+*/
+import envConfig from 'env.config'; // eslint-disable-line import/no-unresolved
```
adamstankiewicz, line 56: [curious] Could we modify the default ESLint rules in `@edx/frontend-build` to disable the `import/no-unresolved` rule specifically for `env.config`? That way, we wouldn't need to disable the ESLint rule here. Given the file is exposed via `@edx/frontend-build`, it feels reasonable to have `@edx/frontend-build` make it pass ESLint, too.

adamstankiewicz (approved):

brian-smith-tcril (approved):
This is awesome! :heart: how well documented everything is!

davidjoy:
So this is up for debate, but I was assuming (and didn't explicitly say anywhere) that env.config.js would be as @arbrandes describes - it wouldn't be checked in, but there'd be an env.config.js.example file that is, and that file would probably contain a helpful comment and a reasonable development/local configuration.  I think that solves the 'private' situation.  env.config.js would be .gitignore'd.  

As for testing, we could add something like this to the `setupTest.js` file:

```
jest.mock('env.config.js', () => ({
  BASE_URL: 'booyah',
  // etc.
}));
```

That would work, and not require us to have a test version of the `env.config.js` file.

The alternative might be to go back to frontend-build and try to make it do something more conditional based on `NODE_ENV` and look for env.config.prod.js, env.config.dev.js, env.config.test.js, env.config.private.js, but I'm not sure that's much better.  The prod config wouldn't be useful for anyone (like .env now), the dev config will likely be different whether you're using Tutor or devstack and only useful for the latter, etc.  Not sure that buys us much.

brian-smith-tcril:
`.gitignore`ing `.env.config.js` sounds like it's probably the move

davidjoy:
Thanks everyone for your input and smarts. 😄  Next up, a DEPR ticket and guidance on how to move forward.

src/initialize.js
adamstankiewicz: @davidjoy See https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md#ignore.

Mashal-m:
Hi @davidjoy! I am currently working on upgrading the [frontend-app-learner-portal-programs](https://github.com/openedx/frontend-app-learner-portal-programs) to React 17. However, I am facing an issue when upgrading the frontend platform to version `4.5.0` (which includes this PR work). I doubt this issue is occurring because the MFE was created without utilizing a frontend build and is using Gatsby. I would like to hear your thoughts on this.
Here is my react 17 [PR](https://github.com/openedx/frontend-app-learner-portal-programs/pull/280). 

Attaching error screenshots below:
Screenshot 2023-07-12 at 12 52 03 PM
Screenshot 2023-07-12 at 12 15 50 PM

adamstankiewicz:
@Mashal-m Note that `@edx/frontend-platform` now has a peer dependency with `@edx/frontend-build`, introduced through this PR. In the current state, it is assumed that `@edx/frontend-build` will be used alongside `@edx/frontend-platform` for MFE applications, which as you pointed out is not the case for `frontend-app-learner-portal-programs`.

See this comment thread for a bit more detail on this coupling with `@edx/frontend-build`: https://github.com/openedx/frontend-platform/pull/474#discussion_r1163324012

From the linked comment thread above: 

> it feels like introducing the peer dependency is potentially a **breaking change** as a result, even though I agree it's probably fair to assume this won't bite anyone.

Looks like our assumption at the time that introducing the coupling between `@edx/frontend-platform` and `@edx/frontend-build` wouldn't be a breaking change for any consumers was unfortunately inaccurate. /cc @davidjoy 

---

Based on your screenshots, the specific issue is importing `env.config.js` within the Jest tests. See `@edx/frontend-build`'s base `jest.config.js` file which configures Jest to handle `env.config.js` imports, for example here where `env.config` is added to `moduleNameMapper`. To resolve your repo's test issues, you might be able to update `frontend-app-learner-portal-programs`' `jest.config.js` file to include a similar config as `@edx/frontend-build`'s.

If you're also running into issues when running/building `frontend-app-learner-portal-programs`, an additional path forward for you may be to update `frontend-app-learner-portal-programs`' `onCreateWebpackConfig` ([source](https://github.com/openedx/frontend-app-learner-portal-programs/blob/26d8c5afc9f6a7dd4766cf2fda44fd900997e4ee/gatsby-node.js#L66)) to replicate what `@edx/frontend-build` does to create a Webpack alias import or, alternatively, possibly use a Gatsby plugin like `gatsby-plugin-alias-imports` ([source](https://www.gatsbyjs.com/plugins/gatsby-plugin-alias-imports/)):

```js
// from https://github.com/openedx/frontend-build/blob/master/config/webpack.common.config.js
// (applied during `npm start` and `npm run build`)

{
  // ....
  resolve: {
    alias: {
      'env.config': path.resolve(process.cwd(), './env.config'),
    },
    fallback: {
      // This causes the system to return an empty object if it can't find an env.config.js file in
      // the application being built.
      'env.config': false,
    },
}
```

davidjoy:
Oof, I think @adamstankiewicz covered it well, so I don't have much to add.  Let us know how it goes.

PR #475, opened 2023-04-12
fix: force refresh jwt token on ensureAuthenticatedUser

syedsajjadkazmii:
[VAN-1371](https://2u-internal.atlassian.net/browse/VAN-1371)

**Description:**

This PR adds a `forceRefresh: true` option while calling `fetchAuthenticatedUser` inside `ensureAuthenticatedUser` to forcefully refresh the JWT token from the backend.

**Why this was needed?**

When a user resets his password through account settings (accounts MFE), the user is redirected to the login page because his session is invalidated at the backend. But he is not logged out of other MFEs such as learner home even if the user refreshes the page. But if the user tries to do some API call to the backend for example user tries to **unenroll** a course in the learner's home MFE, a 403 forbidden response is returned from the backend because he is unauthenticated at the backend and the course is not unenrolled.

The reason is on refreshing the JWT token is not refreshed from the backend instead the token is fetched from cookies and decoded to ensure authentication at the frontend. 

**Effect on users:**

- Users stay authenticated on the frontend but are unable to perform certain actions because they are unauthenticated at the backend.
- This is inconsistent in that the user is unauthenticated at the backend but authenticated at frontend. 

This fix will make sure to un-authenticate the user on frontend when the user is unauthenticated on the backend.

**Please let us know if you think this change will have any negative effects**

### DIFF
--- a/src/auth/AxiosJwtAuthService.js
+++ b/src/auth/AxiosJwtAuthService.js
@@ -250,7 +250,7 @@ class AxiosJwtAuthService {
    * @returns {Promise<UserData>}
    */
   async ensureAuthenticatedUser(redirectUrl = this.config.BASE_URL) {
-    await this.fetchAuthenticatedUser();
+    await this.fetchAuthenticatedUser({ forceRefresh: true });
 
     if (this.getAuthenticatedUser() === null) {
       const isRedirectFromLoginPage = global.document.referrer


syedsajjadkazmii:
> Doesn't this mean we'll be forcing a refresh on every single page load of all MFEs?  This code is called on initialization. 

Yes it does mean that.

adamstankiewicz (commented):
I tend to share David's concerns about force refreshing on initialization of all MFEs. An alternative could be that you override the `auth` handler in the initialization of your MFE in such a way that passes `forceRefresh: true` for just the accounts MFE.

FWIW, we've also had to implement at least one mechanism to force-refresh an authenticated user's JWT token in our Enterprise MFE(s). So, it does seem like `@edx/frontend-platform` should have a re-usable mechanism to allow consumers to force refresh the JWT if a specified condition passes on initialization; I don't think if it should be defaulted to true for all MFEs.

In the Enterprise MFE(s), we ended up implementing a `LoginRefresh` ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/f6a7aecba357f97369828d5c0d7ffb9a6d2d9399/src/components/app/LoginRefresh.jsx#L9)) component that conditionally calls the `${getConfig().LMS_BASE_URL}/login_refresh` API when needed. You might consider something like this as an alternative solution, too.

davidjoy (changes requested):
Based on our conversation, I think we need to have a more nuanced approach here for the particular issue we're running into.  As written, this will cause all MFEs to make an extra request on load, affecting performance across the board.

PR #235, opened 2021-10-21
feat: iframe-based experience plugins

davidjoy:
**Description:**
This PR enables the ability to render plugin MFEs inside a host MFE using Plugin Slots and iFrames. 

**The additions and changes in this PR include:** 
- Plugin setup for both Host (where plugin is rendered) and "Child" MFE (where desired component is retrieved)
- The `example` app was also moved inside its own folder so that an `example-plugin` app could also run concurrently in order to see the plugin framework in use. (31 files)
  - Includes replacing `.env` files with `env.config.js`  

**Recommended Approach to Understanding**
- Run Example Apps
  - Run `npm ci` in both example app directories before running `npm run start` for each app. 
  - Notable files include
    - Host MFE: 
      - `example/src/index`
      - `example/src/PluginsPage`
    - "Child" MFE: 
      - `example-plugin-app/src/index.jsx`
      - `example-plugin-app/src/PluginOne.jsx` (for error fallback)
      - `example-plugin-app/src/PluginTwo.jsx`
- Read ["edX Micro-frontend Plugins Tech Spec"](https://docs.google.com/document/d/1nNxSFSGkGkIQP769WTUN9cSnMImEDEOZFpJ-0qDOR3M) (Restricted to 2U) to gain a better overview of the framework, considerations that still need to be made, and some diagrams. 
- Checkout the branch and follow the flow for:
  - Unit tests written in `Plugin.test.jsx`, include both Host and "Child" 
  - Host MFE
    - PluginSlot -> PluginContainer -> PluginContainerIFrame 
    - includes `hooks`, `constants`, and `pluginConfigShape`
  - "Child" MFE 
    - Plugin + PluginErrorBoundary 

**Post merge:**

- [ ] After the build finishes for the merged commit, verify the new release has been pushed to [NPM](https://www.npmjs.com/package/@edx/frontend-platform).

PR #541, opened 2023-08-18
docs: Add ADR for handling frontend plugin fallback

jsnwesson:
**Description:**

This ADR is to propose how we should handle the scenario where the iFrame plugin fails to render.
The decision here will help with releasing the PR to make iFrame plugins possible in MFEs. (#235)

### DIFF
--- a/docs/decisions/0008-frontend-plugin-fallback.rst
+++ b/docs/decisions/0008-frontend-plugin-fallback.rst
@@ -1,0 +1,38 @@
+How to Handle Frontend iFrame Plugin Fallback 
+======================================================================================
+
+Status
+------
+
+Draft
+
+Context
+-------
+The Frontend Plugin Framework's iFrame class implementation will live in Frontend Platform,
+but the question of how we want to handle the case where the iFrame fails to render needs to
+handled. 
+
+Given that this is a new feature that is being made available to the frontend, we 
+can't anticipate all of the ways that the plugin will be used for a given repo, nor do we 
+have any intention to limit the scope of how it can be used. Because we don't all of the use 
+cases for the plugin, we also don't know the desired fallback method that is desired for any 
+given plugin.
+
+Decision
+--------
+For the above reasons, we will have failing iFrame plugins return an H1 tag that 
+says "There was an error.". By doing so, it is up to the engineer/team that implements the plugin 
+to handle the desired fallback however they would like. 
+
+Rejected Alternatives
+---------------------
+
+The initial assumption was that we should determine what the fallback would be
+if the iFrame plugin render failed. The extent of what this handling could look like 
+ranges from taking in a fallback argument to providing the same fallback image no matter
+how the plugin was being used. These options limit the possibilities for how the fallback 
+-- and the plugin itself -- would present itself by assuming each plugin implementation 
+or desired fallback method will be the same for all plugins.  
+
+.. _oepXXXX: https://github.com/openedx/open-edx-proposals/pull/287
+.. _frontend_plugin_roadmap: https://github.com/openedx/platform-roadmap/issues/27


docs/decisions/0008-frontend-plugin-fallback.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Given that this is a new feature that is being made available to the frontend, we 
+can't anticipate all of the ways that the plugin will be used for a given repo, nor do we 
+have any intention to limit the scope of how it can be used. Because we don't all of the use 
+cases for the plugin, we also don't know the desired fallback method that is desired for any 
+given plugin.
+
+Decision
+--------
+For the above reasons, we will have failing iFrame plugins return an H1 tag that
```
jsnwesson, line 23: I'm fully aware that this is likely not the best way to deliver an error, but in hopes of stirring enough of the pot to bring people into the conversation, I'll leave it here (see [Constructive Controversy](https://www.mindtools.com/absml9b/constructive-controversy))

docs/decisions/0008-frontend-plugin-fallback.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Context
+-------
+The Frontend Plugin Framework's iFrame class implementation will live in Frontend Platform,
+but the question of how we want to handle the case where the iFrame fails to render needs to
+handled. 
+
+Given that this is a new feature that is being made available to the frontend, we 
+can't anticipate all of the ways that the plugin will be used for a given repo, nor do we 
+have any intention to limit the scope of how it can be used. Because we don't all of the use
```
zainab-amir, line 17: nit:
```suggestion
have any intention to limit the scope of how it can be used. Because we don't know all of the use 
```

docs/decisions/0008-frontend-plugin-fallback.rst
```diff
@@ -0,0 +1,38 @@
...
+can't anticipate all of the ways that the plugin will be used for a given repo, nor do we 
+have any intention to limit the scope of how it can be used. Because we don't all of the use 
+cases for the plugin, we also don't know the desired fallback method that is desired for any 
+given plugin.
+
+Decision
+--------
+For the above reasons, we will have failing iFrame plugins return an H1 tag that 
+says "There was an error.". By doing so, it is up to the engineer/team that implements the plugin 
+to handle the desired fallback however they would like.
```
zainab-amir, line 25: How will we do that? By checking for the element returned by the pluginSlot?

docs/decisions/0008-frontend-plugin-fallback.rst
zainab-amir, line 23: Can we log this in a way that the teams can determine how often this fails for their host site?

syedsajjadkazmii: How will the host MFE know that the iFrame has failed to render?

docs/decisions/0008-frontend-plugin-fallback.rst
syedsajjadkazmii: How we will be able to check the element returned by the pluginSlot?

docs/decisions/0008-frontend-plugin-fallback.rst
```diff
@@ -0,0 +1,38 @@
...
+Context
+-------
+The Frontend Plugin Framework's iFrame class implementation will live in Frontend Platform,
+but the question of how we want to handle the case where the iFrame fails to render needs to
+handled. 
+
+Given that this is a new feature that is being made available to the frontend, we 
+can't anticipate all of the ways that the plugin will be used for a given repo, nor do we 
+have any intention to limit the scope of how it can be used. Because we don't all of the use 
+cases for the plugin, we also don't know the desired fallback method that is desired for any
```
adamstankiewicz, line 18:
```suggestion
cases for the plugin, we also don't know the desired fallback method for any 
```

docs/decisions/0008-frontend-plugin-fallback.rst
adamstankiewicz, line 23: [curious] Is there an opportunity to use an `ErrorBoundary` in some way here? Either the one defined/used by `@edx/frontend-platform` or a custom one for this use case.

jsnwesson: Ah! You know, I should update this to name that there is a `PluginErrorBoundary` component that is part of the PR I'm working on for this. 

In short, the `PluginErrorBoundary` will capture any errors that occur within, and whatever decision is made here on how to best reflect an error will be handled by the error boundary.

docs/decisions/0008-frontend-plugin-fallback.rst
jsnwesson: hey all! huge apologies for the delay on these responses. After having written this ADR draft, I realized that the PR I was taking on wasn't in any state for me to confidently know _what worked_ and thus a lack of confidence in the ways we could handle the fallback. 

I will be editing this draft to reflect my findings and we'll ideally have a clearer picture to work off of.

Mashal-m:
Hey, @jsnwesson, could you please pass the failed ci check?

PR #574, opened 2023-10-02
fix: make Tutor-MFE config compatible with React Router v6

ormsbee:
**Description:**

As part of the React Router v5 -> v6 upgrade, history was no longer being passed in. This broke the Tutor deployment method of MFEs, which puts all the MFEs in a common server and uses different top-level dirs to separate them.

The fix for this is to pass in an explicit basename to the Router.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/src/initialize.js
+++ b/src/initialize.js
@@ -103,6 +103,26 @@ export const history = (typeof window !== 'undefined')
   }) : createMemoryHistory();
 
 /**
+ * The string basename that is the root directory of this MFE.
+ *
+ * In devstack, this should always just return an empty string, because each MFE is in its own
+ * server/domain.
+ *
+ * In Tutor, all MFEs are deployed to a common server, each under a different top-level directory.
+ * The basename is the root path for a given MFE, e.g. "/library-authoring". It is set by tutor-mfe
+ * as an ENV variable in the Docker file, and we read it here from that configuration so that it
+ * can be passed into a Router later.
+ *
+ * By convention, our basenames do not have a trailing slash. So it's "/library-authoring", and not
+ * "/library-authoring/". Our configuration is inconsistent on this, so this function always strips
+ * off the trailing slash from any value entered in configuration.
+ */
+export function basename() {
+  const configBasename = getPath(getConfig().PUBLIC_PATH);
+  return configBasename.endsWith('/') ? configBasename.slice(0, -1) : configBasename;
+}
+
+/**
  * The default handler for the initialization lifecycle's `initError` phase.  Logs the error to the
  * LoggingService using `logError`
  *
--- a/src/initialize.js
+++ b/src/initialize.js
@@ -105,22 +105,14 @@ export const history = (typeof window !== 'undefined')
 /**
  * The string basename that is the root directory of this MFE.
  *
- * In devstack, this should always just return an empty string, because each MFE is in its own
- * server/domain.
+ * In devstack, this should always just return "/", because each MFE is in its own server/domain.
  *
  * In Tutor, all MFEs are deployed to a common server, each under a different top-level directory.
  * The basename is the root path for a given MFE, e.g. "/library-authoring". It is set by tutor-mfe
  * as an ENV variable in the Docker file, and we read it here from that configuration so that it
  * can be passed into a Router later.
- *
- * By convention, our basenames do not have a trailing slash. So it's "/library-authoring", and not
- * "/library-authoring/". Our configuration is inconsistent on this, so this function always strips
- * off the trailing slash from any value entered in configuration.
  */
-export function basename() {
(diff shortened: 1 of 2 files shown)


ormsbee:
@brian-smith-tcril, @arbrandes: I'm not sure what needs to be done in terms of tests and version bumping here?

ormsbee:
FYI @kdmccormick

src/initialize.js
```diff
@@ -102,6 +102,26 @@ export const history = (typeof window !== 'undefined')
...
+ * as an ENV variable in the Docker file, and we read it here from that configuration so that it
+ * can be passed into a Router later.
+ *
+ * By convention, our basenames do not have a trailing slash. So it's "/library-authoring", and not
+ * "/library-authoring/". Our configuration is inconsistent on this, so this function always strips
+ * off the trailing slash from any value entered in configuration.
+ */
+export function basename() {
+  const configBasename = getPath(getConfig().PUBLIC_PATH);
+  return configBasename.endsWith('/') ? configBasename.slice(0, -1) : configBasename;
```
adamstankiewicz, line 122: Nit/Note: not all MFEs currently have a `PUBLIC_PATH` configured (e.g., presumably MFEs not officially part of Tutor). That said, things seem to still work as expected in this case as `getConfig().PUBLIC_PATH` falls back to `/` if not provided.

In the case where `getConfig().PUBLIC_PATH` is `/`, this `basename` function returns an empty string. This would force react-router's `Router` component's `basename` prop to fall back to its own `/` instead.

While it seems to be fine as is, I'm wondering if we should be explicitly handle the fallback to `/` on our own here without relying on the fallback within `Router` itself? E.g.

```js
export function basename() {
  const configBasename = getPath(getConfig().PUBLIC_PATH);
  if (configBasename === '/') {
    return configBasename;
  }
  return configBasename.endsWith('/') ? configBasename.slice(0, -1) : configBasename;
}
```

adamstankiewicz (approved):
Aside from one nit/suggestion, these changes LGTM. I tested using frontend-template-application's MFE on top of devstack with the following test cases:

* `PUBLIC_PATH` configured as `/`.
* `PUBLIC_PATH` configured as `/testing`.
* `PUBLIC_PATH` not configured.

All test cases seem to be working as expected.

src/initialize.js
arbrandes: I would avoid stripping the tail `/` here, since:

1. It's not necessary for react-router (as Brian S shows above)
2. It goes beyond feature parity with what worked before the react-router-v6 upgrade

ormsbee: @adamstankiewicz: I'll make it really explicit in that way, thank you.

@brian-smith-tcril: React Router wants to leave it in our control, but I think that _we_ should be consistent and not have a mix of `/library-authoring/`,  `/account`, `/authn/`, etc. Our backend app config is using the ones without trailing slashes, so the current live URLs don't have trailing slashes. So even though I think that having slashes is actually the right thing to do from an academic perspective, I'd rather be standardized on what's already out there.

I do intend to make a separate PR to `tutor-mfe` to fix how it writes this config, but I thought that for rollout/upgrade ease, we should force this behavior across MFEs.

ormsbee: @arbrandes, @brian-smith-tcril: Okay, I want to make sure I'm not misunderstanding something.

Given the current state of tutor-mfe-generated config and the URLs we currently have, the code is broken without that stripping. Tutor MFE code will generate the PUBLIC_PATH as `/library-authoring/`. Trying to load `http://apps.local.overhang.io:3001/library-authoring` (no trailing slash) will cause an error because the route is not specified that way. The only reason this worked before was because React Router used to be more permissive about what it would accept here, and in v6 they've gotten more explicit.

I actually have no objection to removing the stripping behavior and relying on everyone getting the updated tutor-mfe generated config. I was just worried that this would complicate things when people are updating, especially if re-deployment doesn't happen all at once.

ormsbee: I've made the PR to tutor-mfe to fix the config (https://github.com/overhangio/tutor-mfe/pull/154). I'll remove the slash stripping part from this PR.

src/initialize.js
```diff
@@ -89,6 +89,7 @@ import {
   APP_READY, APP_INIT_ERROR,
 } from './constants';
 import configureCache from './auth/LocalForageCache';
+import { config } from 'localforage';
```
ormsbee, line 92: ... a little confused how I even managed to get this in here, but removing shortly....

ormsbee:
@brian-smith-tcril, @arbrandes: Modified to just take the config value as-is, and relying on https://github.com/overhangio/tutor-mfe/pull/154 to make sure we're getting the right value.

arbrandes (approved):
Thanks again, Dave!  Looks good to me!

PR #631, opened 2024-01-04
feat: Add support for Google Tag Manager

xitij2000:
This change adds support for Google Tag Manager along with some common options for Google Tag Manager.

Google Tag Manager is a tag management system that can be used for conversion tracking, analytics etc. This change will allow configuring and using GTM in addition to Google Analytics. 

Private-ref: [BB-8363](https://tasks.opencraft.com/browse/BB-8363)

### DIFF
--- a/src/scripts/GoogleTagManagerLoader.js
+++ b/src/scripts/GoogleTagManagerLoader.js
@@ -1,0 +1,60 @@
+/**
+ * @implements {GoogleTagManagerLoader}
+ * @memberof module:GoogleTagManager
+ */
+class GoogleTagManagerLoader {
+  constructor({ config }) {
+    this.gtmId = config.GOOGLE_TAG_MANAGER_ID;
+    this.gtmArgs = '';
+    if (config.GOOGLE_TAG_MANAGER_AUTH) {
+      this.gtmArgs += `&gtm_auth=${config.GOOGLE_TAG_MANAGER_AUTH}`;
+    }
+    if (config.GOOGLE_TAG_MANAGER_PREVIEW) {
+      this.gtmArgs += `&gtm_preview=${config.GOOGLE_TAG_MANAGER_PREVIEW}`;
+    }
+    if (config.GOOGLE_TAG_MANAGER_ADDNL_ARGS) {
+      if (!config.GOOGLE_TAG_MANAGER_ADDNL_ARGS.startsWith('&')) {
+        this.gtmArgs += '&';
+      }
+      this.gtmArgs += config.GOOGLE_TAG_MANAGER_ADDNL_ARGS;
+    }
+  }
+
+  loadScript() {
+    if (!this.gtmId) {
+      return;
+    }
+
+    global.googleTagManager = global.googleTagManager || [];
+    const { googleTagManager } = global;
+
+    // If the snippet was invoked do nothing.
+    if (googleTagManager.invoked) {
+      return;
+    }
+
+    // Invoked flag, to make sure the snippet is never invoked twice.
+    googleTagManager.invoked = true;
+
+    googleTagManager.load = (key, args, options) => {
+      const scriptSrc = document.createElement('script');
+      scriptSrc.type = 'text/javascript';
+      scriptSrc.async = true;
+      scriptSrc.innerHTML = `
+      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+       new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+       j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+       'https://www.googletagmanager.com/gtm.js?id='+i+dl+'${args}';f.parentNode.insertBefore(j,f);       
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @xitij2000! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

mphilbrick211:
Hi @openedx/fed-bom @openedx/2u-fed-bom! (not sure which group to use). This is ready for review. Thanks!

mphilbrick211:
Hi @openedx/2u-fed-bom! Just following up on this :)

src/scripts/GoogleTagManagerLoader.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * @implements {GoogleTagManagerLoader}
+ * @memberof module:GoogleTagManager
+ */
+class GoogleTagManagerLoader {
+  constructor({ config }) {
+    this.gtmId = config.GOOGLE_TAG_MANAGER_ID;
+    this.gtmArgs = '';
+    if (config.GOOGLE_TAG_MANAGER_AUTH) {
```
abdullahwaheed, line 9: could you please add respective config in the env as well

abdullahwaheed (commented):
could please provide some more context of this change and its impact

src/scripts/GoogleTagManagerLoader.js
xitij2000: I'm not sure where else it needs to be added. It's an optional feature so the it should not be there in the default config.

xitij2000:
> could please provide some more context of this change and its impact

I've updated the PR description, do tell if there is need for additional context or info.

abdullahwaheed (approved):
LGTM

arbrandes (changes requested):
(This started as a conversation in [#wg-frontend](https://openedx.slack.com/archives/C04BM6YC7A6/p1715958731940419).  Transcribing here.)

This looks like it would be useful, so I don't want to get in the way.  At the same time, I think it would make a lot of sense if we could, instead of adding explicit support for GTM in frontend-platform, make it so that people could add any such script they want without having to submit it upstream.  The advantages being:

* For the user: no need to try and get it merged
* For the maintainer: no need to make sure specific providers continue to work, are compliant, etc

I'm thinking that if one could define GoogleTagManagerLoader in an array of such loaders in `env.config.js` with a predefined name... say, `externalScripts`?  Then all we need to do here is make sure that `initialize()` also (or instead) looks for external scripts that come from `env.config`.

Thoughts?

src/scripts/GoogleTagManagerLoader.js
```diff
@@ -0,0 +1,60 @@
...
+  loadScript() {
+    if (!this.gtmId) {
+      return;
+    }
+
+    global.googleTagManager = global.googleTagManager || [];
+    const { googleTagManager } = global;
+
+    // If the snippet was invoked do nothing.
+    if (googleTagManager.invoked) {
```
zawan-ila, line 32: [curious-passerby] Could you give an example where loadScript would be called twice? I feel like it is only being called from initialize, and initialize should only be called once in an app, right?

xitij2000:
> (This started as a conversation in [#wg-frontend](https://openedx.slack.com/archives/C04BM6YC7A6/p1715958731940419). Transcribing here.)
> 
> This looks like it would be useful, so I don't want to get in the way. At the same time, I think it would make a lot of sense if we could, instead of adding explicit support for GTM in frontend-platform, make it so that people could add any such script they want without having to submit it upstream. The advantages being:
> 
>     * For the user: no need to try and get it merged
> 
>     * For the maintainer: no need to make sure specific providers continue to work, are compliant, etc
> 
> 
> I'm thinking that if one could define GoogleTagManagerLoader in an array of such loaders in `env.config.js` with a predefined name... say, `externalScripts`? Then all we need to do here is make sure that `initialize()` also (or instead) looks for external scripts that come from `env.config`.
> 
> Thoughts?

I already replied there but my only issue with reading `externalScripts`  from `env.config.js` is that we can use a plugin mechanism to auto-load it instead. i.e. auto-populate that array with all packages following the pattern `@openedx-plugins/frontend-platform-script-*` . 

Of course, someone can just install the package and manually populate it as well, and that will likely be simpler. I can work on that approach when we have the capacity to continue this work.

arbrandes:
> auto-populate that array with all packages following the pattern @openedx-plugins/frontend-platform-script-*

I think that's a fine idea.  I figure we could support both mechanisms, though.  Auto- and manual loading.

mphilbrick211:
Hi @xitij2000 - just checking in to see if this is still in progress?

xitij2000:
> Hi @xitij2000 - just checking in to see if this is still in progress?

Sorry for the late reply. I'm hoping to work on this in an upcoming sprint.

mphilbrick211:
@xitij2000 can this be closed?

xitij2000:
@Agrendalath I think this should be good to close right? We can implement it via slots now in any case?

Agrendalath:
@xitij2000, yeah, we can close this for now. Before doing this, please link [the ticket](https://tasks.opencraft.com/browse/BB-8363) in the PR's description.

PR #640, opened 2024-01-19
chore: bump and support paragon v22

adamstankiewicz:
**Description:**

Paragon v22 was released. The current peer dependencies in package.json won't allow consumers to upgrade to Paragon v22 until explicit support for the new version is added to package.json.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   },
   "peerDependencies": {
     "@openedx/frontend-build": ">= 13.0.15",
-    "@openedx/paragon": ">= 21.5.7",
+    "@openedx/paragon": ">= 21.5.7 < 23.0.0",
     "prop-types": "^15.7.2",
     "react": "^16.9.0 || ^17.0.0",
     "react-dom": "^16.9.0 || ^17.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
   },
   "peerDependencies": {
     "@openedx/frontend-build": ">= 13.0.15",
-    "@openedx/paragon": "^21.5.7",
+    "@openedx/paragon": ">= 21.5.7",
```
brian-smith-tcril, line 78: are we happy with just a `>=` here? before the switch from `@edx/paragon` to `@openedx/paragon` we had

https://github.com/openedx/frontend-platform/blob/b608ccdb8fff78658b7e6a49afea89a1c91a46c5/package.json#L78

maybe instead we want this to be

```suggestion
    "@openedx/paragon": ">= 21.5.7 < 23.0.0",
```

I don't have strong feelings one way or the other, just wanted to note that this is a departure from patterns we were using previously

adamstankiewicz: Yeah, it would be safer to keep the semver range explicit (with a max version specified); I ended up going with the pattern of `@openedx/frontend-build` above, but agree we should de-risk future Paragon upgrades by making new major versions explicit with the tradeoff being it requires a maintenance a step to bump the version each time Paragon releases a new breaking change (not a big deal, IMO).

For example, if the `Button` component props API were to change in a v23 release and breaks frontend-platform's usage of it, we would want to prevent that by restricting the max version that can be supported.

I will update.

brian-smith-tcril (approved):
:shipit:

PR #689, opened 2024-05-10
feat: adds support for loading external theme CSS for MFEs 

dcoa:
**Description:**

This PR updates the original one https://github.com/openedx/frontend-platform/pull/440 closer to the master branch and adds some extra tests.

Please read the original PR for additional context. 

> [!WARNING]
>  - The dist folder is included in the PR for testing purposes and will be removed before the merge. 
>  - The PR should update the Paragon version once the official release is launched.

**Merge checklist:**

- [x] Consider testing your code modifications in another local micro-frontend using local aliases configured via [the `module.config.js` file in `frontend-build`](https://github.com/openedx/frontend-build#local-module-configuration-for-webpack).

**Post merge:**


openedx-webhooks:
Thanks for the pull request, @dcoa! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

docs/how_tos/theming.md
```diff
@@ -0,0 +1,204 @@
+# Theming support with `@edx/paragon` and `@edx/brand`
```
GlugovGrGlib, line 1: we need to refactor `@edx/` to `@openedx/` in docs, as since December all npm packages are published in the scope of `@openedx/`

GlugovGrGlib: @brian-smith-tcril I couldn't see frontend-platform among npm published packages, shouldn't it be published with @openedx scope as well?
https://www.npmjs.com/search?q=%40openedx

package.json
```diff
@@ -35,14 +35,16 @@
   "devDependencies": {
     "@edx/brand": "npm:@openedx/brand-openedx@^1.2.2",
     "@edx/browserslist-config": "1.2.0",
-    "@openedx/frontend-build": "14.0.3",
-    "@openedx/paragon": "22.3.2",
+    "@openedx/frontend-build": "github:edunext/frontend-build#dcoa/design-tokens-support",
+    "@openedx/paragon": "npm:@openedx/paragon@^22.0.0-alpha.25",
```
GlugovGrGlib, line 39: It might worth to wait for the [Paragon sync PR](https://github.com/openedx/paragon/pull/3044) to be merged, so we won't downgrade paragon requirements, and still will be able to use the latest version of paragon.

PKulkoRaccoonGang:
@dcoa I'm getting an error during build related to the `env.config.js` file

`SyntaxError: Unexpected token 'export'`

I tried to solve it by changing the export to `module.exports = { config };` I succeeded, but the content of the `example` application is not rendered on http://localhost:8080.

Have you encountered such a problem?

docs/how_tos/theming.md
brian-smith-tcril: I think that one is still only published to the `@edx` scope. I agree it would be good to make the switch but it hasn't made it to the top of the priority list yet.

PKulkoRaccoonGang (commented):
@dcoa This looks great! I have already started working on Update `openedx/frontend-app-discussions`. I left a few questions about the operation of some features and the code in general.

docs/how_tos/theming.md
```diff
@@ -0,0 +1,204 @@
...
+}
+```
+
+### Locally installed `@openedx/paragon`
+
+If you would like to use the same version of the Paragon CSS urls as the locally installed `@openedx/paragon`, the configuration for the Paragon CSS urls may contain a wildcard `$paragonVersion` which gets replaced with the locally installed version of `@openedx/paragon` in the consuming application, e.g.:
+
+```shell
+https://cdn.jsdelivr.net/npm/@openedx/paragon@$paragonVersion/dist/core.min.css
+https://cdn.jsdelivr.net/npm/@openedx/paragon@$paragonVersion/dist/light.min.css
```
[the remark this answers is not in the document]
dcoa: No, because the main intention here is to identify the version and load the external resource compatible with that specific version. 

For example, my app has paragon `23.0.0-alpha.2`, then the hook will replace `$paragonVersion` .

This is the definition
`https://cdn.jsdelivr.net/npm/@openedx/paragon@$paragonVersion/dist/core.min.css`

after processing 
`https://cdn.jsdelivr.net/npm/@openedx/paragon@$23.0.0-alpha.2/dist/core.min.css`

docs/how_tos/theming.md
```diff
@@ -0,0 +1,204 @@
...
+
+By serving CSS loaded externally, consuming applications of Paragon no longer need to be responsible for compiling the theme SCSS to CSS themselves and instead use a pre-compiled CSS file. In doing so, this allows making changes to the Paragon theme without needing to necessarily re-build and re-deploy all consuming applications.
+
+### Dark mode and theme variant preferences
+
+`@openedx/frontend-platform` supports both `light` (required) and `dark` (optional) theme variants. The choice of which theme variant should be applied on page load is based on the following preference cascade:
+
+1. **Get theme preference from localStorage.** Supports persisting and loading the user's preference for their selected theme variant, until cleared.
+1. **Detect user system settings.** Rely on the `prefers-color-scheme` media query to detect if the user's system indicates a preference for dark mode. If so, use the default dark theme variant, if one is configured.
```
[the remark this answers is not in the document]
dcoa: The hook will fallback in the installed themes files if the URL is not provided or is not working. 

If you define only the CSS inside SCSS file you can use the `prefers-color-scheme`  to support dark/light modes.

docs/how_tos/theming.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Performance implications
+
+There is also a meaningful improvement in performance as loading the compiled theme CSS from an external CDN means micro-frontends (MFEs) can include cached styles instead of needing to load essentially duplicate theme styles included in each individual MFE as users navigate across the platform.
+
+However, as the styles from `@openedx/paragon` and `@openedx/brand-openedx` get loaded at runtime by `@openedx/frontend-platform`, the associated CSS files do not get processed through the consuming application's Webpack build process (e.g., if the MFE used PurgeCSS or any custom PostCSS plugins specifically for Paragon).
+
+### Falling back to styles installed in consuming application
+
+If any of the configured external `PARAGON_THEME_URLS` fail to load for whatever reason (e.g., CDN is down, URL is incorrectly configured), `@openedx/paragon` will attempt to fallback to the relevant files installed in `node_modules` from the consuming application.
```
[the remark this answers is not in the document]
dcoa: @PKulkoRaccoonGang  thanks for the feedback, the fallback url was missing the `publicPath` I solved it.

image

src/react/hooks/paragon/useParagonThemeUrls.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // If we don't have  the core default or any theme variants, use the PARAGON_THEME
+  if (!coreCss.default || Object.keys(themeVariantsCss).length === 0) {
+    const localCoreUrl = PARAGON_THEME.paragon?.themeUrls?.core;
+    const localThemeVariants = PARAGON_THEME.paragon?.themeUrls?.variants;
+
+    if (!localCoreUrl || Object.keys(localCoreUrl).length === 0
+      || !localThemeVariants || Object.keys(localThemeVariants).length === 0) {
+      return undefined;
+    }
```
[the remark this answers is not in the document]
dcoa: I create the function in the utils file to be used in 2 hooks

src/react/hooks/paragon/useParagonThemeCore.test.js
```diff
@@ -0,0 +1,119 @@
...
+            light: 'light',
+          },
+          variants: {
+            light: {
+              fileName: 'light.min.css',
+            },
+          },
+        },
+      },
+    };
```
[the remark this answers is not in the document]
dcoa: I can get rid of this understanding that is being exported by frontend-build (including for testing) 

**Note:** Once the frontend-build PR has been merged we need to update the `devDependencies` and `peerDependencies`

src/react/hooks/paragon/useParagonThemeUrls.test.js
```diff
@@ -0,0 +1,135 @@
...
+                brandOverride: undefined,
+              },
+            },
+          },
+        }),
+      );
+    });
+
+    it('returns expected object when configuration is valid (both Paragon + brand)', () => {
+      const config = {
```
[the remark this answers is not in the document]
dcoa: I prefer leaving this section as it is. The important part here is `PARAGON_THEME_URLS`, which changes in each test. It's easier to understand this way, I think.

src/react/AppProvider.test.jsx
```diff
@@ -104,4 +140,147 @@ describe('AppProvider', () => {
...
+          <div>Child One</div>
+          <div>Child Two</div>
+        </AppProvider>
+      );
+      render(Component);
+      expect(useTrackColorSchemeChoice).toHaveBeenCalled();
+    });
+
+    it('calls useParagonTheme', () => {
+      const Component = (
```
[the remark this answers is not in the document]
dcoa: I can declare the global component but this test doesn't cover the useParagonTheme behavior so it is not necessary to declare `paragon theme and brand`.  This is focused on the `setThemeVariant` function and the theme state.

docs/how_tos/theming.md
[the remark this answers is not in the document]
PKulkoRaccoonGang: Thank you! It would be great to explicitly indicate exactly how it is possible to use different imports for a light/dark theme in the documentation. I plan to do this in the migration documentation and for Paragon.

For example:
```
@use "@openedx/paragon/dist/core.min.css" as paragonCore;
@import url('@openedx/paragon/dist/light.min.css') (prefers-color-scheme: light);
@import url('@my-brand/dark.min.css') (prefers-color-scheme: dark);
```

docs/how_tos/theming.md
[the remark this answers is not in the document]
PKulkoRaccoonGang: Thanks for the explanation 💯

src/react/hooks/paragon/useParagonThemeUrls.test.js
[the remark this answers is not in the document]
PKulkoRaccoonGang: Good, I agree with you, thanks

src/react/AppProvider.test.jsx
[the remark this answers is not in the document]
PKulkoRaccoonGang: Ok, thanks

docs/how_tos/theming.md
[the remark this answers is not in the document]
dcoa: 100% agree that the information in your example should be part of the Paragon and migration documents.

This is a different approach, the explanation of how to use variants is below, in the title  **Basic theme URL configuration**.

 I will improve one of the examples to use dark mode, to make it more explicit.

docs/how_tos/theming.md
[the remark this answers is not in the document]
PKulkoRaccoonGang: We don’t receive `undefined`, but I couldn’t load styles from the locally installed Paragon package in `node_modules`. Could you check if `getParagonThemeCss` works correctly in a PR related to a [frontend-build](https://github.com/openedx/frontend-build/pull/546)?

image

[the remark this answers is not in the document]
dcoa: The problem here is related to tutor-mfe, they define `BASE_URL` as the `MFE_HOST` ([you can see here](https://github.com/overhangio/tutor-mfe/blob/master/tutormfe/patches/openedx-lms-development-settings#L4)) thing that is not 100% true in dev mode because the URL needs the `PORT` that is specific for each MFE

You can  add in the settings:

``` python
MFE_CONFIG_OVERRIDES = {
     "discussions": {
        "BASE_URL": "apps.local.edly.io:2002"
    }
}
```

package.json
```diff
@@ -76,7 +77,7 @@
   },
   "peerDependencies": {
     "@openedx/frontend-build": ">= 14.0.0",
-    "@openedx/paragon": ">= 21.5.7 < 23.0.0",
+    "@openedx/paragon": " 23.0.0-alpha.1 || >= 21.5.7 < 23.0.0",
```
brian-smith-tcril, line 80: > [!NOTE]  
> Waiting on Paragon alpha to be merged to master

brian-smith-tcril (commented):
Overall this is looking great! I left a suggestion to clarify the documentation is referring to design tokens, and a few comments noting PRs that need to land before this can.

Thank you so much for this!

arbrandes:
@davidjoy, this is a doozy to keep an eye on.

davidjoy:
Sure is.  It looks like it was temporary, but just to explicitly note it: the `dist` folder has been checked in in this PR.  We need to remember to remove that before merging.

package.json
```diff
@@ -35,11 +35,12 @@
   "devDependencies": {
     "@edx/brand": "npm:@openedx/brand-openedx@^1.2.2",
     "@edx/browserslist-config": "1.2.0",
-    "@openedx/frontend-build": "14.0.10",
+    "@openedx/frontend-build": "github:edunext/frontend-build#dcoa/design-tokens-support",
```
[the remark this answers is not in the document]
dcoa: I resolved this understanding frontend-build has been updated and a compatible version (`14.1.5`) is now been used.

xitij2000:
Is there any way I or someone from OpenCraft can help with this? We're using a version of this in production already  and it's working quite well so we'd love to see this merged.

brian-smith-tcril:
Thanks for the interest @xitij2000! It's exciting to hear this is being used already!

At a quick glance things that need to happen for this PR to land are:
* Update `paragon` dependency to allow either v22 or the newly release v23
* Resolve conflicts
* Final QA/Review

dcoa:
Hi @brian-smith-tcril, I updated the peerDependencies (limited to < 24) and rebased the code to master, in that way we can start  the final review :)

brian-smith-tcril:
That's wonderful news @dcoa!

I'm more than happy to start on the code side of the final review here!

Is there still a PR to an MFE (using Paragon 22) somewhere? Having a sandbox up that verifies this won't break anything when it lands makes merging this a lot less scary!

dcoa:
@brian-smith-tcril we have this one https://github.com/openedx/frontend-app-discussions/pull/731 but is not updated, if you can help me with that or if you prefer I can open a new one.

brian-smith-tcril:
> @brian-smith-tcril we have this one [openedx/frontend-app-discussions#731](https://github.com/openedx/frontend-app-discussions/pull/731) but is not updated, if you can help me with that or if you prefer I can open a new one.

That works! I'd be happy to update that one (or open a new one if updating it becomes a pain)

src/react/hooks/paragon/utils.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export const handleVersionSubstitution = ({ url, wildcardKeyword, localVersion }) => {
+  if (!url || !url.includes(wildcardKeyword) || !localVersion) {
+    return url;
+  }
+  return url.replaceAll(wildcardKeyword, localVersion);
+};
+
+export const isEmptyObject = (obj) => !obj || Object.keys(obj).length === 0;
```
brian-smith-tcril, line 101: It looks like someone did some investigation into performance of different ways to test for empty objects in js (https://stackoverflow.com/a/59787784) and found that using for-in is the fastest

 ```suggestion
export const isEmptyObject = (obj) => {
  for (var i in obj) {
    return false;
  }
  return true;
}
```

I don't think this is a required change at all, more just something fun I found when looking into it!

**Review by brian-smith-tcril — COMMENTED** [review]
Overall this is looking awesome! Super happy to see all the puzzle pieces coming together!

I left a few comments in there, mostly just notes in places where naming/comments don't make what's happening entirely clear.

Thank you so much for this!

File: src/react/hooks/paragon/useParagonThemeUrls.js [hunk 13]
```diff
@@ -0,0 +1,90 @@
...
+    return undefined;
+  }
+  const paragonThemeUrls = config.PARAGON_THEME_URLS;
+  const paragonCoreCssUrl = typeof paragonThemeUrls?.core?.urls === 'object' ? paragonThemeUrls.core.urls.default : paragonThemeUrls?.core?.url;
+  const brandCoreCssUrl = typeof paragonThemeUrls?.core?.urls === 'object' ? paragonThemeUrls.core.urls.brandOverride : undefined;
+  const defaultThemeVariants = paragonThemeUrls.defaults;
+
+  // Local versions of @openedx/paragon and @edx/brand
+  const localParagonVersion = PARAGON_THEME?.paragon?.version;
+  const localBrandVersion = PARAGON_THEME?.brand?.version;
```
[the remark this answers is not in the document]
dcoa: what do you thing about moving the function `handleVersionSubstitution` into this file? 
Despite it does not change the checking place, might facilitate the code reading.

.env.development
```diff
@@ -17,7 +17,7 @@ STUDIO_BASE_URL=http://localhost:18010
 MARKETING_SITE_BASE_URL=http://localhost:18000
 ORDER_HISTORY_URL=http://localhost:1996/orders
 REFRESH_ACCESS_TOKEN_ENDPOINT=http://localhost:18000/login_refresh
-SEGMENT_KEY=''
```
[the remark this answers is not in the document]
dcoa: No need it, it was a refactor to keep all the undefined variables same, I reverted it

env.config.js
```diff
@@ -5,4 +5,4 @@ const config = {
   JS_FILE_VAR: 'JS_FILE_VAR_VALUE_FOR_EXAMPLE_APP',
 };
 
-export default config;
+module.exports = config;
```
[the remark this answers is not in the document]
dcoa: We made that changes because we encountered some issues with webpack (when we update frontend-build here with the PR code) ,  I tested it again and is worked so I reverted the change.

webpack.dev.config.js
```diff
@@ -14,4 +14,9 @@ module.exports = createConfig('webpack-dev', {
       '@edx/frontend-platform': path.resolve(__dirname, 'src'),
     },
   },
+  devServer: {
+    static: {
+      directory: path.resolve(__dirname, 'example/dist'),
+    },
+  },
```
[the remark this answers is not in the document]
dcoa: same comment as `env.config.js`

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,212 @@
...
+  themeVariants,
+  currentThemeVariant,
+  onLoad,
+  onDarkModeSystemPreferenceChange,
+}) => {
+  const [isParagonThemeVariantLoaded, setIsParagonThemeVariantLoaded] = useState(false);
+  const [isBrandThemeVariantLoaded, setIsBrandThemeVariantLoaded] = useState(false);
+
+  useEffect(() => {
+    const someFn = (colorSchemeQuery) => {
```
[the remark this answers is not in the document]
dcoa: I updated the name to `changeColorShceme`, let me know what do you think.

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,212 @@
...
+              setIsParagonThemeVariantLoaded(true);
+            }
+            const otherExistingLinks = getExistingThemeVariantLinks(isBrandOverride);
+            removeExistingLinks(otherExistingLinks);
+            return;
+          }
+          const paragonThemeAccessor = isBrandOverride ? 'brand' : 'paragon';
+          const themeUrls = PARAGON_THEME?.[paragonThemeAccessor]?.themeUrls ?? {};
+          if (themeUrls.variants && themeUrls.variants[themeVariant]) {
+            const themeVariantFallbackUrl = fallbackThemeUrl(themeUrls.variants[themeVariant].fileName);
```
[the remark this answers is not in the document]
dcoa: Good catch!,  I changed that here and in `useParagonThemeCore`

src/react/hooks/paragon/useParagonThemeUrls.js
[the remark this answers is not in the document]
brian-smith-tcril: > what do you thing about moving the function `handleVersionSubstitution` into this file? Despite it does not change the checking place, might facilitate the code reading.

That sounds great to me! In general I like to avoid putting code in a `utils` file whenever possible. Keeping helpers near the code they're helping is perfect!

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,212 @@
...
+    const colorSchemeQuery = window.matchMedia?.('(prefers-color-scheme: dark)');
+    if (colorSchemeQuery) {
+      colorSchemeQuery.addEventListener('change', changeColorShceme);
+    }
+    return () => {
+      if (colorSchemeQuery) {
+        colorSchemeQuery.removeEventListener('change', changeColorShceme);
+      }
+    };
+  }, [onDarkModeSystemPreferenceChange]);
```
brian-smith-tcril, line 42: This name is much better than `someFn`! Looks like a slight typo made it in though.

```suggestion
  useEffect(() => {
    const changeColorScheme = (colorSchemeQuery) => {
      onDarkModeSystemPreferenceChange(colorSchemeQuery.matches);
    };
    const colorSchemeQuery = window.matchMedia?.('(prefers-color-scheme: dark)');
    if (colorSchemeQuery) {
      colorSchemeQuery.addEventListener('change', changeColorScheme);
    }
    return () => {
      if (colorSchemeQuery) {
        colorSchemeQuery.removeEventListener('change', changeColorScheme);
      }
    };
  }, [onDarkModeSystemPreferenceChange]);
```

src/config.js
```diff
@@ -171,6 +171,7 @@ let config = {
   MFE_CONFIG_API_URL: process.env.MFE_CONFIG_API_URL,
   APP_ID: process.env.APP_ID,
   SUPPORT_URL: process.env.SUPPORT_URL,
+  PARAGON_THEME_URLS: process.env.PARAGON_THEME_URLS ? JSON.parse(process.env.PARAGON_THEME_URLS) : {},
```
adamstankiewicz, line 174: If the environment variable `PARAGON_THEME_URLS` has invalid JSON, is there a chance this `JSON.parse(...)` would throw an error? Might it be worth implementing a helper function to wrap the `JSON.parse(...)` with a `try / catch` for improved error handling?

src/react/hooks/paragon/useParagonTheme.js
```diff
@@ -0,0 +1,111 @@
...
+ * current theme variant's CSS. This ensures that if the theme variant is changed at runtime, the CSS for the new theme
+ * variant will already be loaded.
+ *
+ * @memberof module:React
+ * @param {object} config An object containing the URLs for the theme's core CSS and any theme (i.e., `getConfig()`)
+ *
+ * @returns An array containing 2 elements: 1) an object containing the app
+ *  theme state, and 2) a dispatch function to mutate the app theme state.
+ */
+const useParagonTheme = (config) => {
```
adamstankiewicz, line 28: [nit/consideration] Likely could simplify the usage of this hook slightly, as I believe `getConfig` could be called directly within this hook, where appropriate, versus needing to pass the `config` object as a function parameter.

For example, would it make sense to call `getConfig` within `useParagonThemeUrls` directly instead of passing `config` variable around?

src/react/AppProvider.jsx
```diff
@@ -63,7 +68,27 @@ export default function AppProvider({ store, children, wrapWithRouter }) {
...
+
+        // Persist selected theme variant to localStorage.
+        window.localStorage.setItem(SELECTED_THEME_VARIANT_KEY, themeVariant);
+      },
+    },
+  }), [authenticatedUser, config, locale, paragonThemeState, paragonThemeDispatch]);
+
+  if (!paragonThemeState?.isThemeLoaded) {
+    return null;
+  }
```
adamstankiewicz, line 91: [sanity check] Given when this code path is executed, the entire MFE app is rendered as a blank white screen, I'm wondering if we have sufficient observability for when we return `null` here, after any actual async loading has finished/resolved.

For example, if there's an issue with the CDN and we also fail to fallback to using the CSS source files from the locally installed `@openedx/paragon` and `@edx/brand` packages, and the MFE returns a blank white screen, will we sufficiently get alerted to it via `logError`, etc.?

src/react/hooks/paragon/useParagonThemeCore.js
```diff
@@ -0,0 +1,144 @@
...
+}) => {
+  const [isParagonThemeCoreLoaded, setIsParagonThemeCoreLoaded] = useState(false);
+  const [isBrandThemeCoreLoaded, setIsBrandThemeCoreLoaded] = useState(false);
+
+  useEffect(() => {
+    // Call `onLoad` once both the paragon and brand theme core are loaded.
+    if (isParagonThemeCoreLoaded && isBrandThemeCoreLoaded) {
+      onLoad();
+    }
+  }, [isParagonThemeCoreLoaded, isBrandThemeCoreLoaded, onLoad]);
```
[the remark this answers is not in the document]
adamstankiewicz: Would something like `resolve` or `complete` be more generically applicable to the cases of legit loading vs the other scenarios "loading" was set truthy?

For example
* `setIsParagonThemeCoreResolved` and `onResolve`
* `setIsParagonThemeCoreComplete` and `onComplete`

Rationale: indicates that the async process is "done" vs. "loaded".

src/react/hooks/paragon/useParagonThemeCore.js
```diff
@@ -0,0 +1,144 @@
...
+          if (isBrandOverride) {
+            setIsBrandThemeCoreLoaded(true);
+          } else {
+            setIsParagonThemeCoreLoaded(true);
+          }
+        };
+        coreThemeLink.onerror = () => {
+          logError(`Failed to load core theme CSS from ${url}`);
+          if (isFallbackThemeUrl) {
+            logError(`Could not load core theme CSS from ${url} or fallback URL. Aborting.`);
```
adamstankiewicz, line 69: [curious] Any concerns around having two `logError` function calls within the same code path? I.e., if `isFallbackThemeUrl` is truthy here, would we have logged basically the same error twice (with slightly differing error messages)?

src/react/hooks/paragon/useParagonThemeCore.js
```diff
@@ -0,0 +1,144 @@
...
+            if (isBrandOverride) {
+              setIsBrandThemeCoreLoaded(true);
+            } else {
+              setIsParagonThemeCoreLoaded(true);
```
adamstankiewicz, line 73: [curious] Do we want to ensure coverage on the critical code paths within `useParagonThemeCore`, e.g. around loading? There's a few CodeCov annotations about missing coverage.

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,212 @@
...
+            } else {
+              setIsParagonThemeVariantLoaded(true);
+            }
+          }
+        };
+
+        themeVariantLink.onerror = () => {
+          logError(`Failed to load theme variant (${themeVariant}) CSS from ${value.urls.default}`);
+          if (isFallbackThemeUrl) {
+            logError(`Could not load theme variant (${themeVariant}) CSS from fallback URL. Aborting.`);
```
adamstankiewicz, line 119: [curious] Similar question as above, regarding potential for calling `logError` twice for largely the same error, when `isFallbackThemeUrl` is truthy. Should we only call `logError` once in this case?

src/react/hooks/paragon/useParagonThemeVariants.js
```diff
@@ -0,0 +1,212 @@
...
+            if (isBrandOverride) {
```
adamstankiewicz, line 120: Similar question around test coverage on some of the uncovered code paths within this hook.

src/react/index.js
```diff
@@ -13,4 +13,5 @@ export { default as ErrorBoundary } from './ErrorBoundary';
 export { default as ErrorPage } from './ErrorPage';
 export { default as LoginRedirect } from './LoginRedirect';
 export { default as PageWrap } from './PageWrap';
-export { useAppEvent } from './hooks';
+export { useAppEvent, useParagonTheme } from './hooks';
+export { paragonThemeActions } from './reducers';
```
adamstankiewicz, line 17: [sanity check / clarification] Can you expand on why we've want to export  `paragonThemeActions` to consumers? I believe the intent for consumers to manage switching theme variants is to use the `paragonTheme.setThemeVariant` function exposed within `AppContext`?

When will consumers use the exported `paragonThemeActions`?

src/config.js
brian-smith-tcril: That sounds like a great plan. then this line could just be
```js
PARAGON_THEME_URLS: parseParagonThemeUrls(process.env.PARAGON_THEME_URLS),
```

and the function could handle all the logic (both `process.env.PARAGON_THEME_URLS` being falsy and invalid json)

src/react/hooks/paragon/useParagonTheme.js
dcoa: Yeah, I think this make sense because the configuration variables tend to be loaded at the first load (ideally once) and not change during the user navigation. I will made the suggested change.

src/config.js
```diff
@@ -137,6 +137,16 @@ function extractRegex(envVar) {
   return undefined;
 }
 
+function parseParagonThemeUrls(paragonUrlsJson) {
```
brian-smith-tcril, line 140: I think it makes sense to check for a falsy value before trying to parse. Not having `process.env.PARAGON_THEME_URLS` set is not an error state, but should still return `{}`.

Then, assuming we make it past the falsy check, we can go into the `try/catch`. It would also be good to log errors that we catch in there.

I think specifically checking for `err` being a `SyntaxError` exception (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions) would be good. So something like

```js
  try {
    return JSON.parse(paragonUrlsJson);
  } catch (err) {
    if (e instanceof SyntaxError) {
      // log something about not being able to parse the json and it possibly being invalid.
      // maybe also log something about what to expect themes to look like when {} is returned
      return {};
    } else {
      // not sure what to do here, if we're getting a different error type maybe we should just throw it?
    }
  }
```

src/react/hooks/paragon/useParagonThemeCore.js
dcoa: Yes, working on the overall coverage :)

src/react/AppProvider.jsx
dcoa: I think we have enough paths for loading potential errors during the loading, I will working on the testing that will help to cover that cases and potentially identify if some validation is missing

src/react/hooks/paragon/useParagonThemeCore.js
dcoa: Should we just left the `isFallbackThemeUrl` and relay on the fallback logInfo as the can not load the default URL?

https://github.com/openedx/frontend-platform/pull/689/files#diff-6e8530492255f6f52120a1dd9f69147538c94d6e37be13541f802a2189f80af8R83

package.json
```diff
@@ -36,13 +36,15 @@
     "@edx/brand": "npm:@openedx/brand-openedx@^1.2.2",
     "@edx/browserslist-config": "1.4.0",
     "@openedx/frontend-build": "14.2.2",
-    "@openedx/paragon": "22.13.0",
+    "@openedx/paragon": "23.1.1",
```
PKulkoRaccoonGang, line 39: [question]: Does it make sense to provision the latest minor version instead of a hard-coded version?

```suggestion
    "@openedx/paragon": "^23.1.1",
```

src/config.js
```diff
@@ -137,6 +137,19 @@ function extractRegex(envVar) {
...
```
PKulkoRaccoonGang, line 140: [optional/question]: Since we are not using TypeScript in this file, does it make sense to add [JSDocs](https://jsdoc.app/about-getting-started) instead of comments that will contain both the description of the function and some types for `paragonUrlsJson`? I think this approach will make the code easier to read, better to understand, and will simplify the potential transition to TypeScript in the future.

This question applies to all new functions in this PR.

dist/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### One-time setup if you have not upgraded node/npm
+IMPORTANT: There is now a new node/npm version being used by frontend-platform as of
+https://github.com/openedx/frontend-platform/pull/259
+
+#### Install nvm
+This is highly recommended to be able to leverage different node/npm versions.
+For a some time, different repositories may be using different versions of node/npm.
+
+Alternatively, please install node16 and npm8 for use with this repository.
```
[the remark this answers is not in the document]
brian-smith-tcril: It looks like that's currently a problem in the `README` on `master`.

https://github.com/openedx/frontend-platform/blob/15ef507e41127b4fd4ace5d31f7e527381678572/README.md?plain=1#L36

I think we should address that in a separate PR.

src/config.js
dcoa: [curious] Currently `{}`  will not configure any URL (similar behavior as `undefined`), do we expect (prefer) a different behavior (for example, try to load the default themes)?

src/config.js
dcoa: I add the JSDoc for the mentioned function and enhance the other ones.

package.json
dcoa: I made the change :)

docs/how_tos/theming.md
```diff
@@ -0,0 +1,250 @@
+# Theming support with `@openedx/paragon` and `@openedx/brand-openedx`
```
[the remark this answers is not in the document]
brian-smith-tcril: It looks like the formatting on this is a little bit broken
image

I think adding some newlines should clean it up

docs/how_tos/theming.md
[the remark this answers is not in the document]
PKulkoRaccoonGang: Resolved

docs/how_tos/theming.md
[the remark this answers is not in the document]
PKulkoRaccoonGang: Resolved

src/react/hooks/paragon/useParagonThemeCore.test.js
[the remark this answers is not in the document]
PKulkoRaccoonGang: Resolved

src/react/hooks/paragon/useParagonThemeUrls.test.js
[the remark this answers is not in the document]
PKulkoRaccoonGang: Resolved

src/react/AppProvider.test.jsx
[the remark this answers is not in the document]
PKulkoRaccoonGang: Resolved

src/config.js
PKulkoRaccoonGang: Resolved

PKulkoRaccoonGang:
@brian-smith-tcril I would love to, but I don’t have such a button. I think it’s because I don’t have the necessary permissions in this repository.

PKulkoRaccoonGang:
@dcoa You can close all the comments I wrote where I added "Resolved" :)

brian-smith-tcril:
@PKulkoRaccoonGang sorry about that! Thanks for commenting on the ones that have been resolved! I went through and clicked the button on them.

src/react/AppProvider.jsx
brian-smith-tcril: From discussion during Paragon WG:
* logging in `useEffect` to avoid duplicate logs
* differentiate between error state and loading state 
* @brian-smith-tcril to look into what logs show up in the console for existing error states to get a sense of a reasonable logging strategy for this
* possible future improvement (once React 18 support lands) - use suspense?

src/react/index.js
brian-smith-tcril: @brian-smith-tcril to look into if this has been updated and resolve

src/config.js
brian-smith-tcril: Discussed in Paragon WG: Following the same pattern as the CDN URLs being down (loading locally installed styles) seems like the best path forward here.

| `PARAGON_THEME_URLS` | Behavior |
| - | - |
| Doesn't exist (MFE choosing to use local styles) | Load local styles (from MFE's style entrypoint) |
| Set, parses properly | Attempt to load from parsed URL(s), if error - fallback to local styles |
| Set, fails to parse | Fallback to local styles (injected from `frontend-platform` - derived from node modules) |

src/react/index.js
brian-smith-tcril: looks like this was addressed in https://github.com/openedx/frontend-platform/commit/e3211d9a7200dc7fe63cec5992ccea3522f30fdb

src/react/hooks/paragon/useParagonThemeCore.js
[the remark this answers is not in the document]
dcoa: my bad (I was confused)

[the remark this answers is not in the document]
brian-smith-tcril: Resolved in https://github.com/openedx/frontend-platform/pull/689/commits/5db22d0bb8cae9afa96457c106430c724cd8e975 - Thank you!

src/react/AppProvider.jsx
brian-smith-tcril: I tried removing everything in the `.env.development` file in `frontend-app-learning` that exists in the `MFE_CONFIG` when running tutor dev. I verified the MFE works as expected when running `npm run dev` using the proper `MFE_CONFIG_API_URL`. I then changed that command to point to `localghost` instead of `localhost` and the MFE loaded with the following error state:

Screenshot From 2025-02-24 09-05-06

Here's the diff I used:
```diff
bsmith@aximdev:~/code/frontend-app-learning$ git diff
diff --git a/.env.development b/.env.development
index 56541aa..78219f7 100644
--- a/.env.development
+++ b/.env.development
@@ -3,31 +3,16 @@
 
 NODE_ENV='development'
 
-ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload'
 APP_ID='learning'
-BASE_URL='http://localhost:2000'
 CONTACT_URL='http://localhost:18000/contact'
-CREDENTIALS_BASE_URL='http://localhost:18150'
 CREDIT_HELP_LINK_URL='https://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/SFD_credit_courses.html#keep-track-of-credit-requirements'
-CSRF_TOKEN_API_PATH='/csrf/api/v1/token'
-DISCOVERY_API_BASE_URL='http://localhost:18381'
-DISCUSSIONS_MFE_BASE_URL='http://localhost:2002'
 ECOMMERCE_BASE_URL='http://localhost:18130'
 ENABLE_JUMPNAV='true'
 ENABLE_NOTICES=''
 ENTERPRISE_LEARNER_PORTAL_HOSTNAME='localhost:8734'
 EXAMS_BASE_URL=''
-FAVICON_URL=https://edx-cdn.org/v3/default/favicon.ico
 IGNORED_ERROR_REGEX=''
-LANGUAGE_PREFERENCE_COOKIE_NAME='openedx-language-preference'
-LMS_BASE_URL='http://localhost:18000'
-LOGIN_URL='http://localhost:18000/login'
-LOGOUT_URL='http://localhost:18000/logout'
-LOGO_URL=https://edx-cdn.org/v3/default/logo.svg
-LOGO_TRADEMARK_URL=https://edx-cdn.org/v3/default/logo-trademark.svg
-LOGO_WHITE_URL=https://edx-cdn.org/v3/default/logo-white.svg
 LEGACY_THEME_NAME=''
-MARKETING_SITE_BASE_URL='http://localhost:18000'
 ORDER_HISTORY_URL='http://localhost:1996/orders'
 PORT=2000
 PROCTORED_EXAM_FAQ_URL=''
@@ -35,9 +20,7 @@ PROCTORED_EXAM_RULES_URL=''
 REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
 SEARCH_CATALOG_URL='http://localhost:18000/courses'
 SEGMENT_KEY=''
-SITE_NAME='edX'
 SOCIAL_UTM_MILESTONE_CAMPAIGN='edxmilestone'
-STUDIO_BASE_URL='http://localhost:18010'
 SUPPORT_URL='https://support.edx.org'
 SUPPORT_URL_CALCULATOR_MATH='https://support.edx.org/hc/en-us/articles/360000038428-Entering-math-expressions-in-assignments-or-the-calculator'
 SUPPORT_URL_ID_VERIFICATION='https://support.edx.org/hc/en-us/articles/206503858-How-do-I-verify-my-identity'
@@ -45,7 +28,6 @@ SUPPORT_URL_VERIFIED_CERTIFICATE='https://support.edx.org/hc/en-us/articles/2065
 TERMS_OF_SERVICE_URL='https://www.edx.org/edx-terms-service'
 TWITTER_HASHTAG='myedxjourney'
 TWITTER_URL='https://twitter.com/edXOnline'
-USER_INFO_COOKIE_NAME='edx-user-info'
 SESSION_COOKIE_DOMAIN='localhost'
 CHAT_RESPONSE_URL='http://localhost:18000/api/learning_assistant/v1/course_id'
 PRIVACY_POLICY_URL='http://localhost:18000/privacy'
diff --git a/package.json b/package.json
index 92532a3..0bd294d 100644
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "postinstall": "patch-package",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
-    "dev": "PUBLIC_PATH=/learning/ MFE_CONFIG_API_URL='http://localhost:8000/api/mfe_config/v1' fedx-scripts webpack-dev-server --progress --host apps.local.openedx.io",
+    "dev": "PUBLIC_PATH=/learning/ MFE_CONFIG_API_URL='http://localghost:8000/api/mfe_config/v1' fedx-scripts webpack-dev-server --progress --host apps.local.openedx.io",
     "test": "fedx-scripts jest --coverage --passWithNoTests",
     "test:watch": "fedx-scripts jest --watch --passWithNoTests",
     "types": "tsc --noEmit"
```

src/config.js
dcoa: I will made some changes to apply the expected behavior thanks

src/react/hooks/paragon/useParagonThemeUrls.js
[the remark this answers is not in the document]
dcoa: implemented with commit https://github.com/openedx/frontend-platform/pull/689/commits/54c6e00474b1b6d4bfb2094c74b31ff2c2ef8329

src/config.js
brian-smith-tcril: I'm just trying to wrap my head around the logic here. From what I can tell:
* `parseParagonThemeUrls` is only ever used when getting `PARAGON_THEME_URLS` from `process.env`, not when getting it from the config API or when using js based config.
  * This makes sense. When getting `PARAGON_THEME_URLS` from the config API or from js based config no json parsing is needed.
* When `getConfig` is called, assuming only `process.env` config is being used
  * If there was nothing in `process.env.PARAGON_THEME_URLS` then `PARAGON_THEME_URLS` will be `undefined`
  * If there was something in `process.env.PARAGON_THEME_URLS`, but the json fails to parse, a syntax error will be logged and `PARAGON_THEME_URLS` will be `{}`
  * If there was some other error, then `PARAGON_THEME_URLS` will be that error object

___

I also didn't directly respond to
> [curious] Currently `{}`  will not configure any URL (similar behavior as `undefined`), do we expect (prefer) a different behavior (for example, try to load the default themes)?

My response to that is: Because this function is only used when parsing the json from `process.env`, it is not a place where any logic related to loading defaults should live. This matches up with how this function is currently written, so we should be good.

___

This leaves me with a few questions about this function specifically.
* Why use `undefined` in one scenario and `{}` in the other? If these are treated differently when `PARAGON_THEME_URLS` is pulled via `getConfig` later on, it'd be good to add some comments explaining why each is used in each scenario.
* Does using `undefined` in once scenario and `{}` complicate using `MFE_CONFIG_OVERRIDES`?  I getting overrides from the config API should be fine with either, but I figured it's worth asking.

src/react/hooks/paragon/useParagonThemeCore.js
```diff
@@ -0,0 +1,150 @@
...
+    // If there is no config for the core theme url, do nothing.
+    if (!themeCore?.urls) {
+      setIsParagonThemeCoreComplete(true);
+      setIsBrandThemeCoreComplete(true);
+      return;
+    }
+    const getParagonThemeCoreLink = () => document.head.querySelector('link[data-paragon-theme-core="true"]');
+    const existingCoreThemeLink = document.head.querySelector(`link[href='${themeCore.urls.default}']`);
+    const brandCoreLink = document.head.querySelector(`link[href='${themeCore.urls.brandOverride}']`);
+    if (!existingCoreThemeLink) {
```
brian-smith-tcril, line 46: This is a bit of a style nit, but having so much nested in this `if` statement is making this a bit hard to follow for me. Would inverting the condition and using an early return work here?

Something like:
```js
if (existingCoreThemeLink) {
  existingCoreThemeLink.rel = 'stylesheet';
  existingCoreThemeLink.removeAttribute('as');
  existingCoreThemeLink.dataset.paragonThemeCore = true;
  if (brandCoreLink) {
    brandCoreLink.rel = 'stylesheet';
    brandCoreLink.removeAttribute('as');
    brandCoreLink.dataset.brandThemeCore = true;
  }
  setIsParagonThemeCoreComplete(true);
  setIsBrandThemeCoreComplete(true);
  return;
}

const getExistingCoreThemeLinks = [ ... ]
```

src/react/hooks/paragon/useParagonThemeCore.js
brian-smith-tcril: > We can have a first message (in the `logInfo` - we also can replace it with a `logError`) that is 
**_Could not load core theme CSS from ${url}. Falling back to locally installed core theme CSS: ${coreThemeFallbackUrl}_**
> 
> and then with the `isFallbackThemeUrl` the second message
**_Could not load core theme fallback URL. Aborting._**
> 
> What do you think?

This sounds/looks perfect to me. Thank you so much for the update!

src/react/hooks/paragon/useParagonThemeVariants.js
brian-smith-tcril: I'm 100% on board with the solutions implemented in https://github.com/openedx/frontend-platform/commit/040ca83c3aae3917e6e5e56b119af77d196c414f and  https://github.com/openedx/frontend-platform/commit/a62e8db758fdabba2ccf1f010151ff79c095b96d!

brian-smith-tcril:
We have the following testing PRs with sandboxes verifying functionality
* https://github.com/openedx/frontend-app-discussions/pull/754
* https://github.com/openedx/frontend-app-discussions/pull/723
* https://github.com/openedx/frontend-app-discussions/pull/724
* https://github.com/openedx/frontend-app-discussions/pull/726

I created a new sandbox for the verify non-breaking PR today. Everything is still working as expected as of writing.

@PKulkoRaccoonGang if you'd like to update your PRs to verify the latest code from this branch still works for you that would be good, but if you feel confident in the changes made since you deployed your PRs then I'm OK with leaving those as-is.

Since this PR has grown to 41 commits and nearly 130 comments, I'd like to take inventory of the state of reviews and split up feedback into 2 categories: "must be addressed before this PR can land" and "can be handled in a follow-up PR"

I would categorize all of my remaining feedback on this PR (other than "don't check in `dist`) as "can be handled in a follow-up." Once `dist` is removed from this branch I will be happy to approve this.

src/react/hooks/paragon/useParagonThemeCore.js
dcoa: Updated here https://github.com/openedx/frontend-platform/pull/689/commits/b5069bafc06a3ffc4321310acfd8727478c08952

PKulkoRaccoonGang:
> @PKulkoRaccoonGang if you'd like to update your PRs to verify the latest code from this branch still works for you that would be good, but if you feel confident in the changes made since you deployed your PR sandboxes then I'm OK with leaving those as-is.

@brian-smith-tcril Thanks! These PRs definitely need to be updated and tested. I'll check with Christian if I can do that.

src/react/hooks/paragon/useParagonThemeVariants.js
brian-smith-tcril: @brian-smith-tcril to double check to see if this has been updated

PKulkoRaccoonGang:
> @brian-smith-tcril Thanks! These PRs definitely need to be updated and tested. I'll check with Christian if I can do that.

@brian-smith-tcril @dcoa I updated the PRs for MFE Discussions, and the PRs for the header and footer were also rebased.

**MFE Discussions**
- [feat: MFE runtime configuration API](https://github.com/openedx/frontend-app-discussions/pull/724)
- [feat: Using Paragon CSS External Hosting (JavaScript-based configuration)](https://github.com/openedx/frontend-app-discussions/pull/726)
- [feat: Using locally installed Paragon CSS](https://github.com/openedx/frontend-app-discussions/pull/723)

- [frontend-component-footer](https://github.com/openedx/frontend-component-footer/pull/303)
- [frontend-component-header](https://github.com/openedx/frontend-component-header/pull/351)

brian-smith-tcril:
@PKulkoRaccoonGang I just did some quick testing on the sandboxes and it's looking great!

| PR | Test post in discussions MFE on sandbox |
| - | - |
| https://github.com/openedx/frontend-app-discussions/pull/723 (locally installed) | https://app.pr-723-6b242e.sandboxes.opencraft.hosting/discussions/course-v1:test+12+2024/my-posts/1 |
| https://github.com/openedx/frontend-app-discussions/pull/724 (runtime config) | https://app.pr-724-6b242e.sandboxes.opencraft.hosting/discussions/course-v1:test+111+2024/my-posts/1 |
| https://github.com/openedx/frontend-app-discussions/pull/726 (js config/CDN) | https://app.pr-726-6b242e.sandboxes.opencraft.hosting/discussions/course-v1:test+1212+2024/my-posts/1 |

@PKulkoRaccoonGang - is there any other testing you'd like to do while this still has `dist` checked in? If not, I think it's time to get `dist` back into `.gitignore` and get ready to land this!

brian-smith-tcril:
@dcoa in https://github.com/openedx/frontend-platform/pull/689#discussion_r1970200348 I mentioned wondering
> Why use undefined in one scenario and {} in the other? If these are treated differently when PARAGON_THEME_URLS is pulled via getConfig later on, it'd be good to add some comments explaining why each is used in each scenario.

I didn't see anything that made me think that `undefined` and `{}` are being handled differently. Is there something I'm missing?

If I'm missing something and `undefined` and `{}` *are* being handled differently, could you add some comments to `parseParagonThemeUrls` explaining that?

If I'm *not* missing anything, could we update `parseParagonThemeUrls` to return `undefined` in both cases?

src/config.js
dcoa: > Why use undefined in one scenario and {} in the other? If these are treated differently when PARAGON_THEME_URLS is pulled via getConfig later on, it'd be good to add some comments explaining why each is used in each scenario.

Meanwhile `undefined`   do not configure the urls in any case, having `{}`   does (fallback urls), the main reason, from my perspective. `useParagonThemeURL`  needs to difference those 2 cases. 

> Does using undefined in once scenario and {} complicate using MFE_CONFIG_OVERRIDES? I assume getting overrides from the config API should be fine with either, but I figured it's worth asking.

The behavior from the API will be the same, if the user configures a `{}`   will fallback in the local files, or if do not configure it or want to remove the variable using `undefined` will do nothing.

adamstankiewicz: > if the user configures a {} will fallback in the local files, or if do not configure it or want to remove the variable using undefined will do nothing.

@dcoa @brian-smith-tcril This distinction seems reasonable to me!

brian-smith-tcril (approved):
Thank you so much for being incredibly helpful and patient throughout the long review process for this PR!

The code here is in stellar shape, and I can't wait to land this!

I noticed the GH web UI says there are conflicts that must be resolved. I checked and it seems those are purely `package.json`/`package-lock.json` changes. Once those are resolved this should be good to land!

Thanks again!

src/react/hooks/paragon/useParagonTheme.test.js
```diff
@@ -0,0 +1,177 @@
...
+  },
+  variants: {
+    light: {
+      urls: {
+        default: 'https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0/dist/light.min.css',
+      },
+    },
+    dark: {
+      urls: {
+        default: 'https://cdn.jsdelivr.net/npm/@edx/paragon@21.0.0/dist/dark.min.css',
```
adamstankiewicz, line 27: [non-blocking nit] Should these CDN urls point to `@openedx/paragon@23` at this point? These files technically don't exist on v21.

src/react/hooks/paragon/useParagonThemeUrls.test.js
```diff
@@ -0,0 +1,217 @@
...
+          },
+          defaults: {
+            light: 'light',
+          },
+          variants: {},
+        },
+      };
+      const originalParagonTheme = global.PARAGON_THEME;
+      Object.defineProperty(global, 'PARAGON_THEME', {
+        value: 'mocked-theme',
```
adamstankiewicz, line 158: [non-blocking nit]: would it make sense to override `PARAGON_THEME` here with an `undefined` vs. string `mocked-theme`? I feel it'd be slightly more realistic (i.e., `PARAGON_THEME` not defined vs. defined with an incorrect value).

src/react/hooks/paragon/useParagonThemeUrls.test.js
```diff
@@ -0,0 +1,217 @@
...
+          },
+        },
+      };
+      mergeConfig(config);
+      const { result } = renderHook(() => useParagonThemeUrls(config));
+      expect(result.current).toEqual(
+        expect.objectContaining({
+          core: {
+            urls: {
+              default: 'https://cdn.jsdelivr.net/npm/@edx/paragon@1.0.0/dist/core.min.css',
```
adamstankiewicz, line 198: [non-blocking nit]: Should this test suite define `PARAGON_THEME` on its own vs. relying on the original provided (presumably from the `setupTest.js` in `@openedx/frontend-build`)? For instance, if the version defined in frontend-build's `setupTest.js` changes to `2.0.0`, I believe this test suite would begin failing.

Explicitly (re-)defining `PARAGON_THEME` with a static version in this test suite could make sense to prevent this from happening 🤔

src/react/hooks/paragon/useParagonThemeVariants.js
adamstankiewicz: This conversation thread can be resolved (I no longer have permissions to do so myself).

adamstankiewicz (approved):
LGTM, with just a handful of non-blocking nits mostly impacting tests! Fantastic job working through the review iterations! 🙌

xitij2000:
Great to see this merge!  🥳

PR #692, opened 2024-05-14
build: upgrade codecov from v3 to v4, use secret

kdmccormick:
Per https://openedx.atlassian.net/wiki/spaces/COMM/pages/3438280709/Adding+Codecov, I think this is the right thing to do.

@dcoa , could you review this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,4 +39,7 @@ jobs:
         run: npm run docs
 
       - name: Run Coverage
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
+        with:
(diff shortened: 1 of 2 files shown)


kdmccormick:
Hm, the Run Coverage step says:
```
Error: Codecov token not found. Please provide Codecov token with -t flag.
```

dcoa:
There is a second file that uses codecov too, [manual-publish.yml ](https://github.com/openedx/frontend-platform/blob/master/.github/workflows/manual-publish.yml#L29)

kdmccormick:
@dcoa how does this look?

dcoa (approved):
Looks good to me, thanks @kdmccormick

PR #701, opened 2024-06-03
chore: replace 'global' with 'globalThis'

bradenmacdonald:
The `global` keyword is a non-standard NodeJS specific keyword. But since 2019, we can use `globalThis` instead, which works in every runtime. For example, `globalThis` has the same consistent meaning in both browsers and NodeJS, whereas `global` does not work in browsers unless you use some build process to rewrite it to `window`.

Details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#description

Example from Chrome console:
Screenshot of chrome browser console


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

bradenmacdonald:
Note: tests are passing but the CI encountered the codecov token limit error.

adamstankiewicz (approved):
Seems reasonable. LGTM!

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-platform&entry.867055334=https://github.com/openedx/frontend-platform/pull/701&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2024-06-03+18:27&entry.1917517419=2024-06-18+17:08&entry.2133058324=Yes) so we can improve your experience in the future.

.eslintrc.js
```diff
@@ -3,6 +3,7 @@ const { getBaseConfig } = require('@openedx/frontend-build');
 
 const config = getBaseConfig('eslint');
 
+config.globals = { globalThis: 'readonly' };
```
adamstankiewicz, line 6: @bradenmacdonald [curious] Lots of repos use `global` today; I'm wondering if we should update the base ESLint config defined in `@openedx/frontend-build` so each consuming shared JS library and/or MFE doesn't need to specify this config across many different repos. Curious your two cents.

On a similar note, I wonder if it'd be more acceptable to extend the shared ESLint config's `env` option to include `es2020` so that other similar syntax is deemed acceptable as well? `@openedx/frontend-build` relies on the `env` defined in `@edx/eslint-config` [here](https://github.com/openedx/eslint-config/blob/89d0227e2d1569b7b9cdf8a29e37fbb3e5600c12/.eslintrc.js#L20-L24). Setting `es2020` in `env` would in place of configuring the `globals` for `globalThis` explicitly.

bradenmacdonald: Yes, I'm definitely in favor of setting `es2020` in `env` as a default!

adamstankiewicz: @bradenmacdonald Opened a PR to the base shared ESLint configuration (`@edx/eslint-config`): https://github.com/openedx/eslint-config/pull/153

I believe having `es2020` in there would eliminate the need to keep `es6`, since the environments (as far as I can tell) are cumulative/additive, where a later environment supersedes an earlier one. That said, I'm opting to keep `es6` in there for now just in case that is not true.

Let me know your thoughts; will also share in FWG Slack :)

PR #723, opened 2024-08-09
feat: support ability for consumers to add custom props / HTML attributes defined via configuration to components

adamstankiewicz:
# Description

## Context

Open edX instances (e.g., 2U) rely on third-party tools for logging and analytics (e.g., Datadog, Hotjar). Some of these tools end up ingesting PII into things such as [session replays](https://www.datadoghq.com/product/real-user-monitoring/session-replay/), etc. that should otherwise be masked.

Example vendor-specific approaches to masking UI elements from their ingested data:
* [**Datadog**](https://docs.datadoghq.com/real_user_monitoring/session_replay/browser/privacy_options/#override-an-html-element). Supports `data-dd-privacy="allow" | "mask" | "hidden" | "mask-user-input"`.
* [**Hotjar**](https://help.hotjar.com/hc/en-us/articles/115012439167-How-to-Suppress-Text-Images-Videos-and-User-Input-from-Collected-Data#specific_elements). Supports `data-hj-suppress` and `data-hj-allow`.
* [**Fullstory**](https://help.fullstory.com/hc/en-us/articles/360020623574-How-do-I-protect-my-users-privacy-in-Fullstory#code-first) Supports `.fs-mask`, `.fs-unmask`, `.fs-exclude`, etc. class names.
* Other third-party vendors, with their own specific HTML attributes/values/classes.

Currently, such attributes for Hotjar are hardcoded throughout the platform ([search results](https://github.com/search?q=org%3Aopenedx%20data-hj-suppress&type=code)), despite not all instances of Open edX using Hotjar. While these attributes are fairly benign, the current approach does not support extending the masking of PII to other vendors (e.g., Datadog).

Related, there are other vendor-specific HTML attributes and class names unrelated to PII this approach would also be applicable to. As such, the following implementation is generic to support "custom props" on annotated elements, regardless of whether it's for PII masking or otherwise.

## Solution

Introduces a React hook (`useComponentPropOverrides`) and a Higher-Order-Component (`withComponentPropOverrides `) as a mechanism to allow consumers to add one or more custom props (e.g., to mask PII within [session replays](https://www.datadoghq.com/product/real-user-monitoring/session-replay/)) to any component that accepts prop spreading (e.g., `...props`).

### Configuration

May be configured by `env.config.js` or the MFE runtime configuration with `componentPropOverrides` (open to other config property names):

```js
{
  componentPropOverrides: {
    targets: {
      username: {
        'data-dd-privacy': 'mask', // Custom `data-*` attribute (e.g., Datadog)
        'data-hj-suppress': '', // Custom `data-*` attribute (e.g., Hotjar)
        className: 'fs-mask', // Custom `className` attribute (e.g., Fullstory)
      },
    },
  },
}
```

### Usage

Consider the following JSX containing PII (i.e., username):

```jsx
<p>Hi there, {authenticatedUser.username}.</p>
```

#### `useComponentPropOverrides` (hook)

```jsx
function UsernameWithPropOverrides({ children, ...rest }) {
  const propOverrides = useComponentPropOverrides('username', rest);
  return <span {...propOverrides}>{children}</span>;
}
UsernameWithPropOverrides.propTypes = {
  children: PropTypes.node.isRequired,
};

export default function AuthenticatedPage() {
  const { authenticatedUser, config } = useContext(AppContext);
  return (
      <p>Hi there, <UsernameWithPropOverrides>{authenticatedUser.username}</UsernameWithPropOverrides>.</p>
  );
}
```

#### `withComponentPropOverrides` (HOC)

```jsx
function Username({ children, ...rest }) {
  return <span {...rest}>{children}</span>;
}
Username.propTypes = {
  children: PropTypes.node.isRequired,
};
const UsernameWithPropOverrides = withComponentPropOverrides('username')(Username);

export default function AuthenticatedPage() {
  const { authenticatedUser, config } = useContext(AppContext);
  return (
      <p>Hi there, <UsernameWithPropOverrides>{authenticatedUser.username}</UsernameWithPropOverrides>.</p>
  );
}
```

By using `useComponentPropOverrides` (hook) or `withComponentPropOverrides` (HOC) to mask the username from third-party tools with the above `componentPropOverrides` configuration, the resulting HTML would render as the following:

```html
<p>Hi there, <span data-dd-privacy="mask" data-hj-suppress="" class="fs-mask">edx</span>.</p>
```

It does this by attempting to find a match between the component's specified selector and the `componentPropOverrides.targets.*` defined via configuration. If the specified selector matches one of the selectors defined via configuration, any configured props and their values be returned.

## Special cases

By default, components supporting configurable prop overrides only works with the prop `className` and any prop prefixed with `data-` (e.g., `data-dd-privacy`). Any other configuration prop name is ignored and not applied during rendering.

In certain cases, components may opt-in to supporting overrides of explicitly named prop names (e.g., `onClick`). By making the special cases like function handlers or `style` opt-in within the MFE's base code, it provides opportunity to discuss/review which prop overrides are officially supported in any given component beyond the default `data-*` attributes or `className` prop.

### `className`

Custom class name(s) will be concatenated with any existing `className` prop values that might already be passed to the component.

### `style`

Custom `style` properties will be shallow merged with any existing `style` properties that might already exist for the component. If the style properties overlap, the custom `style`'s value takes precedence.

### Functions (e.g., `onClick`)

If a custom prop is defined as a function (i.e., supported when `customComponentProps` is defined via `env.config`) but the component itself already has an `onClick` function handler, the component's `onClick` function will be executed first before calling the custom `onClick` handler. Both functions otherwise receive the same functions `args`.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/docs/template/edx/publish.js
+++ b/docs/template/edx/publish.js
@@ -629,7 +629,7 @@ exports.publish = (memberData, opts, tutorials) => {
         generateSourceFiles(sourceFiles, opts.encoding);
     }
 
-    // if (members.globals.length) { generate('Global', [{kind: 'globalobj'}], globalUrl); }
+    if (members.globals.length) { generate('Global', [{kind: 'globalobj'}], globalUrl); }
 
     // index page displays information from package.json and lists files
     files = find({kind: 'file'});
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
 const { getBaseConfig } = require('@openedx/frontend-build');
 
 const config = getBaseConfig('eslint');
--- a/env.config.js
+++ b/env.config.js
@@ -1,8 +1,41 @@
-// NOTE: This file is used by the example app.  frontend-build expects the file
+/* eslint-disable no-console */
+
+// NOTE: This file is used by the example app. frontend-build expects the file
 // to be in the root of the repository.  This is not used by the actual frontend-platform library.
-// Also note that in an actual application this file would be added to .gitignore.
+// Also note that in an actual application, this file would be added to .gitignore.
 const config = {
   JS_FILE_VAR: 'JS_FILE_VAR_VALUE_FOR_EXAMPLE_APP',
+  componentPropOverrides: {
+    targets: {
+      example: {
+        'data-dd-privacy': 'mask', // Custom `data-*` attribute (e.g., Datadog)
+        'data-hj-suppress': '', // Custom `data-*` attribute (e.g., Hotjar)
+        className: 'fs-mask', // Custom `className` attribute (e.g., Fullstory)
+        onClick: (e) => { // Custom `onClick` attribute
+          console.log('[env.config] onClick event for example', e);
+        },
+        style: { // Custom `style` attribute
+          background: 'blue',
+          color: 'white',
+        },
+      },
+      example2: {
+        'data-dd-privacy': 'mask', // Custom `data-*` attribute (e.g., Datadog)
+        'data-hj-suppress': '', // Custom `data-*` attribute (e.g., Hotjar)
+        className: 'fs-mask', // Custom `className` attribute (e.g., Fullstory)
+        onClick: (e) => { // Custom `onClick` attribute
+          console.log('[env.config] onClick event for example2', e);
+        },
+        style: { // Custom `style` attribute
+          background: 'blue',
+          color: 'white',
+        },
+      },
+      example3: {
+        'data-dd-action-name': 'example name', // Custom `data-*` attribute (e.g., Datadog)
+      },
+    },
+  },
 };
 
 export default config;
--- a/example/ComponentPropOverridesPage.jsx
+++ b/example/ComponentPropOverridesPage.jsx
@@ -1,0 +1,152 @@
+import {
+  forwardRef, useContext, useEffect, useRef, useState,
+} from 'react';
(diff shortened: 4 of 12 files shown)


example/PiiPage.jsx
```diff
@@ -0,0 +1,71 @@
...
+        <UsernameWithPii>
+          {authenticatedUser.username}
+        </UsernameWithPii>
+      </p>
+
+      {/* Example 2 */}
+      <h3><code>withPii</code> (HOC)</h3>
+      <p>
+        <i>Username:</i>{' '}
+        <UsernameWithPii2 ref={usernameRef}>
```
adamstankiewicz, line 65: [inform] Shows example passing a `ref` that should be forwarded.

docs/template/edx/publish.js
```diff
@@ -629,7 +629,7 @@ exports.publish = (memberData, opts, tutorials) => {
         generateSourceFiles(sourceFiles, opts.encoding);
     }
 
-    // if (members.globals.length) { generate('Global', [{kind: 'globalobj'}], globalUrl); }
+    if (members.globals.length) { generate('Global', [{kind: 'globalobj'}], globalUrl); }
```
adamstankiewicz, line 632: [inform] generates the `global.html` file which is used in links for custom `@typedef` types used within the React module. without this, the generated JSDoc link results in a 404.

adamstankiewicz:
Closed in favor of https://github.com/openedx/frontend-plugin-framework/pull/84

PR #790, opened 2025-04-04
fix: treat @openedx/frontend-build as an optional peerDep to avoid direct dep in consuming MFEs

adamstankiewicz:
**Related discussions:**
* https://github.com/openedx/frontend-platform/pull/787
* https://github.com/openedx/open-edx-proposals/pull/615

**Corresponding PR:**
* https://github.com/openedx/frontend-app-communications/pull/221

**Description:**

Demonstrates treating `@openedx/frontend-build` as an **optional** peer dependency within `@edx/frontend-platform`, such that consuming MFEs no longer considers `@openedx/frontend-build` as a direct dependency, via:

```json
{
  "peerDependenciesMeta": {
      "@openedx/frontend-build": {
        "optional": true
      }
    }
}
```

From @brian-smith-tcril [here](https://github.com/openedx/frontend-platform/pull/787#issuecomment-2775706607):

> As for `frontend-build` specifically, the fact that it is currently in `frontend-platform`'s `peerDependencies` means that MFEs adding `frontend-platform` as a direct dependency end up with a non-dev dependency on `frontend-build`. In `frontend-app-communications` for example, `@testing-library/react` is in the package lock [with `"dev": true`](https://github.com/openedx/frontend-app-communications/blob/1cd02a9dfb85ddc9e594b6e4b490f33e26a7cac1/package-lock.json#L4895), while `frontend-build` (which is in `package.json` [as a dev dep](https://github.com/openedx/frontend-app-communications/blob/1cd02a9dfb85ddc9e594b6e4b490f33e26a7cac1/package.json#L70)) is [not in `package-lock.json` as dev](https://github.com/openedx/frontend-app-communications/blob/1cd02a9dfb85ddc9e594b6e4b490f33e26a7cac1/package-lock.json#L3707).

This indeed does pose a problem for consuming MFEs as `@openedx/frontend-build` should not be considered as a direct `dependencies` when the MFE itself installs `@openedx/frontend-build` as a `devDependencies`.

Why might this optional peer dependency might work?

> Npm will not automatically install optional peer dependencies ([docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta))

Testing the optional peer dependency approach with frontend-app-communications ([PR](https://github.com/openedx/frontend-app-communications/pull/221)), `@openedx/frontend-build` is no longer treated a direct dependency, but instead receives `devOptional: true` in `package-lock.json` ([source](https://github.com/adamstankiewicz/openedx-frontend-app-communications/blob/72bdd540aa0c4d4caf9cc171cb31be8f83228ccd/package-lock.json#L3767)).

While `npm ci --omit=dev` doesn't omit `@openedx/frontend-build` from being installed for the MFE, `npm ci --omit=dev --omit=optional` would.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "@edx/frontend-platform",
-  "version": "1.0.0-semantically-released",
+  "name": "@adamstankiewicz/openedx-frontend-platform",
+  "version": "8.3.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "@edx/frontend-platform",
-      "version": "1.0.0-semantically-released",
+      "name": "@adamstankiewicz/openedx-frontend-platform",
+      "version": "8.3.6",
       "license": "AGPL-3.0",
       "dependencies": {
         "@cospired/i18n-iso-languages": "4.2.0",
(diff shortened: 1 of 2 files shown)


brian-smith-tcril (approved):
Let's go for it!

PR #876, opened 2026-04-15
feat: make externalScripts configurable via env.config.js

### LINKED ISSUE #877: Make externalScripts configurable via env.config.js
### Description

Currently, `GoogleAnalyticsLoader` is hardcoded as the default value for the `externalScripts` parameter in `initialize()`. Unlike `loggingService`, `analyticsService`, and `authService`, there is no way to configure `externalScripts` via `env.config.js` or the runtime config API.

This means consumers cannot control which external scripts are loaded without passing the option directly to `initialize()`, which is not always feasible depending on how the MFE is set up.

The `externalScripts` option should follow the same config resolution pattern as the other services: check `getConfig().externalScripts` first, then fall back to the parameter default.

### TODO

- [x] https://github.com/openedx/frontend-platform/pull/876
- [x] https://github.com/openedx/public-engineering/issues/505
- [x] https://github.com/overhangio/tutor-mfe/pull/290

arbrandes:
### Description

Currently, `externalScripts` in `initialize()` cannot be configured via `env.config.js`, unlike other services such as `loggingService`, `analyticsService`, and `authService`.

This change makes `externalScripts` resolvable from the config object, following the same `getConfig().externalScripts || externalScripts` pattern already used for the other services. The `loadExternalScripts` call is moved after config resolution so the config-based value is available. `GoogleAnalyticsLoader` remains the default.

Operators who want to customize the list can do so in their `env.config.js`:

```js
import { GoogleAnalyticsLoader } from '@edx/frontend-platform/scripts';
import { CustomScript } from './scripts';

const config = {
  externalScripts: [GoogleAnalyticsLoader, CustomScript],
};

export default config;
```

Closes #877.

### LLM usage notice

Built with assistance from Claude.

### DIFF
--- a/src/initialize.js
+++ b/src/initialize.js
@@ -270,7 +270,7 @@ function applyOverrideHandlers(overrides) {
  * implementation to use.
  * @param {*} [options.authMiddleware=[]] An array of middleware to apply to http clients in the auth service.
  * @param {*} [options.externalScripts=[GoogleAnalyticsLoader]] An array of externalScripts.
- * By default added GoogleAnalyticsLoader.
+ * Can also be configured via `externalScripts` in env.config.js.
  * @param {*} [options.requireAuthenticatedUser=false] If true, turns on automatic login
  * redirection for unauthenticated users.  Defaults to false, meaning that by default the
  * application will allow anonymous/unauthenticated sessions.
@@ -308,10 +308,6 @@ export async function initialize({
     await runtimeConfig();
     publish(APP_CONFIG_INITIALIZED);
 
-    loadExternalScripts(externalScripts, {
-      config: getConfig(),
-    });
-
     // This allows us to replace the implementations of the logging, analytics, and auth services
     // based on keys in the ConfigDocument.  The JavaScript File Configuration method is the only
     // one capable of supplying an alternative implementation since it can import other modules.
@@ -320,6 +316,11 @@ export async function initialize({
     const loggingServiceImpl = getConfig().loggingService || loggingService;
     const analyticsServiceImpl = getConfig().analyticsService || analyticsService;
     const authServiceImpl = getConfig().authService || authService;
+    const externalScriptsImpl = getConfig().externalScripts || externalScripts;
+
+    loadExternalScripts(externalScriptsImpl, {
+      config: getConfig(),
+    });
 
     // Logging
     configureLogging(loggingServiceImpl, {
--- a/src/initialize.test.js
+++ b/src/initialize.test.js
@@ -10,7 +10,7 @@ import {
   APP_READY,
   APP_INIT_ERROR,
 } from './constants';
-import { initialize } from './initialize';
+import { initialize, loadExternalScripts } from './initialize';
 import { subscribe } from './pubSub';
 
 import {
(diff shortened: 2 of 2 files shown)


brian-smith-tcril:
While this definitely feels like an improvement over the current state of things, I'm wondering if it makes the "happy path" for adding simple scripts too verbose.

It seems with this if someone wanted to "just add a script" they'd need to create an entire `Loader` class (or function that does the equivalent) for it because `loadExternalScripts` uses `new`.

https://github.com/openedx/frontend-platform/blob/44cc02429404ed3547c20abd834faf4e487d2c00/src/initialize.js#L201-L206

This means the simplest case for adding `<script src="https://example.com/foo.js">` would be

```jsx
class FooLoader {
  loadScript() {
    const script = document.createElement('script');
    script.src = 'https://example.com/foo.js';
    document.head.appendChild(script);
  }
}
```

I would guess the site operator expectation would be that just adding "`https://example.com/foo.js`" to `externalScripts` would work.

In that case we would probably want to implement/support some sort of generic loader.

_____

The other thing that stands out to me is the `config: getConfig()` in
https://github.com/openedx/frontend-platform/blob/11e94d09b52e358631134aefc36b63b5c7b27480/src/initialize.js#L320-L322

I realize that was already there before this PR, and is currently used in

https://github.com/openedx/frontend-platform/blob/44cc02429404ed3547c20abd834faf4e487d2c00/src/scripts/GoogleAnalyticsLoader.js#L6-L8

but it generally feels like odd coupling to me. Is the expectation that site operators will just add arbitrary things to `MFE_CONFIG` for their external scripts? I know we technically support that in the `ConfigDocument` interface:

https://github.com/openedx/frontend-platform/blob/44cc02429404ed3547c20abd834faf4e487d2c00/src/config.ts#L232

but it still feels a little odd.

mboisson:
> This means the simplest case for adding `<script src="https://example.com/foo.js">` would be
> 
> ```js
> class FooLoader {
>   loadScript() {
>     const script = document.createElement('script');
>     script.src = 'https://example.com/foo.js';
>     document.head.appendChild(script);
>   }
> }
> ```
> 
> I would guess the site operator expectation would be that just adding "`https://example.com/foo.js`" to `externalScripts` would work.
> 
> In that case we would probably want to implement/support some sort of generic loader.

I definitely agree with that, although some configurability is useful. For example, to install the "CookieYes" banner, the code is: 
```
<script id="cookieyes" type="text/javascript" src="https://cdn-cookieyes.com/client_data/REDACTED/script.js"></script>
``` 
i.e. there's the "type" and the "id" that is specified. Not sure if they are needed or only recommended practices.

arbrandes:
@brian-smith-tcril and @mboisson 

> I would guess the site operator expectation would be that just adding "https://example.com/foo.js" to externalScripts would work.

This would not be flexible enough for all use cases.   See Google Analytics itself.

> Is the expectation that site operators will just add arbitrary things to MFE_CONFIG for their external scripts?

Yes.  If you're consuming a loader somebody else published (like the Google Analytics one), it needs to be configurable, somehow.

> For example, to install the "CookieYes" banner, the code is:

Precisely.

mboisson:
There could still be a generic loader class with some level of configurability, enough to handle most cases.

arbrandes:
@mboisson,

> a generic loader class with some level of configurability, enough to handle most cases.

"Some" level of configurability would just be guesswork at this point, and no configurability beyond a URL would not be useful at all.

I believe we should offer the most general solution possible - this one, for example :) - and then once people start using it more, and if they report their use cases, we can figure out what the "minimum common denominator" is for a generic loader.

(Though at that point we'll probably be in frontend-base land entirely, so the solution will look like an app, as opposed to a loader.)

brian-smith-tcril:
> @brian-smith-tcril and @mboisson
> 
> > I would guess the site operator expectation would be that just adding "https://example.com/foo.js" to externalScripts would work.
> 
> This would not be flexible enough for all use cases. See Google Analytics itself.

I 100% agree. I wasn't suggesting we only support the "just a URL" case, but instead support both the URL (via a generic loader) and custom loaders.

That being said, it seems that "just a URL" isn't really a use case. It seems most scripts people want to add need some sort of config, so supporting  "just a URL" seems like something we don't really need.

_____

I would like to dig into the "some level of configurability" idea a bit. A generic loader could just take an object and make a script tag with exactly those attributes on it (assuming they're valid to put on a script tag).

So for the
```html
<script id="cookieyes" type="text/javascript" src="https://cdn-cookieyes.com/client_data/REDACTED/script.js"></script>
```
example the object could be
```js
{ src: 'https://cdn-cookieyes.com/client_data/REDACTED/script.js', id: 'cookieyes', type: 'text/javascript' }
```

_____

One main concern I have about relying on site operators making custom loaders stems from a pain point I've seen with FPF. Site operators want to use shared components in FPF without needing to publish those to NPM. Looking at the README in https://github.com/overhangio/tutor-mfe/pull/290 it seems the recommended way for site operators to add scripts would require publishing a loader to NPM, so I'd expect that to be a pain point here too.

I feel like it'd be pretty great if site operators could just add

```py
  EXTERNAL_SCRIPTS.add_items([
      ("all", "{ src: 'https://cdn-cookieyes.com/client_data/REDACTED/script.js', id: 'cookieyes', type: 'text/javascript' }"),
  ])
```

I'm very open to looking into other ways to mitigate that pain point. Maybe that looks like publishing some common loaders to NPM (or finding people throughout the community to publish/maintain them). Maybe that looks like publishing a generic loader to NPM and having a reasonable config shape for that loader (that isn't tied to `frontend-platform` or `frontend-base`).

Overall this PR is a huge improvement over the current "`externalScripts` is just hardcoded GA" state of things, but I'd like to make sure we think through the site operator experience a bit more before landing this.

brian-smith-tcril:
Ok, after discussing this/thinking this through I agree that landing this PR as-is is the right call.

The one remaining concern I have is that this changes what is required of site operators to enable GA (moving from just setting `GOOGLE_ANALYTICS_4_ID` as documented [here](https://docs.openedx.org/en/latest/site_ops/how-tos/google-analytics.html) to needing to create build-time config)

I think that would require a DEPR (which I think we should make!), but for this PR just keeping it in the build by default is likely the safest option.

arbrandes:
@brian-smith-tcril, agreed: I left GA in, so the behavior is exactly the same if people do nothing.

mboisson:
> One main concern I have about relying on site operators making custom loaders stems from a pain point I've seen with FPF. Site operators want to use shared components in FPF without needing to publish those to NPM. Looking at the README in [overhangio/tutor-mfe#290](https://github.com/overhangio/tutor-mfe/pull/290) it seems the recommended way for site operators to add scripts would require publishing a loader to NPM, so I'd expect that to be a pain point here too.


Reading the above PR, it looks like the easiest way would not be to publish to NPM, but to rather implement a build time definition patch with the loader's code. A generic loader published on NPM would make it easier for all users, but once you're already configuring stuff with tutor's hooks and patches, it is not too foreign to do it with a build time patch.

arbrandes:
@mboisson 

> to rather implement a build time definition patch with the loader's code

Exactly.  I just added an example of this [to the README](https://github.com/overhangio/tutor-mfe/pull/290/changes#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR516).  (Used Cooki👀, because why not.)

brian-smith-tcril (approved):
:shipit:

